PR #5, opened 2019-12-02
Create Dockerfile

mKay00:
I created a dockerfile which installs all necessary build tools and clones the repository. I used VS Code to auto-generate the file and added the install and clone steps at the end. There is no build initiated, since I don't know what's the best utilization for the dockerfile (e.g. to be able to run fiction or for development or both). Based on the use case, the file could be cleaned up (delete the vs code specific commands), a smaller image could be used (e.g. alpine) and a clean-up after the build could be done.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM alpine:latest
 
+# Optional argument to run the "make" command in parallel with the specified NUMBER_OF_JOBS
+ARG NUMBER_OF_JOBS=1
+
 # Configure apt and install packages
     RUN apk add --no-cache \
     # Install cmake with its dependencies
@@ -19,7 +22,7 @@ WORKDIR /fiction
 RUN mkdir build \
     && cd build \
     && cmake .. \
-    && make
+    && make -j${NUMBER_OF_JOBS}
 
 # Automatically start fiction when started in interactive mode
 CMD ["./build/fiction"]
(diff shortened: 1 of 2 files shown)


mKay00:
I updated the dockerfile to use alpine linux, which results in a smaller footprint. Additionally I added build steps, so fiction gets automatically built and I added a step so that fiction gets automatically started, when the image is run.

Additionally I updated the README.md with a section for the usage with a Dockerfile.

marcelwa:
Great additions! Thank you very much for your effort.

Can you add the `-j` option to the `make` command in the `Dockerfile` using the maximum amount of threads available to the image? This would speed up the build significantly.

mKay00:
If I use "make -j", I get build errors and docker crashes. With "make -j4" the build succeeds, so I guess -j starts too many jobs and I have either not enough cores or not enough memory assigned to docker. It is likely, that other users might have the same issue, so I added an optional build argument to docker so the user can select the desired number of jobs. I also updated the readme accordingly.

marcelwa:
That is exactly what I hoped to see. Thanks for your nice work. Will merge as soon as the checks have passed.

mKay00:
Great! 

But the documentation has an issue:

>where <#> is the desired maximum number of jobs. Which value is prudent to use depends on your system and the resources allocated by Docker. Reasonable values that should run on most systems without problems are values between 2 and 8. If your build fails, try using a lower value or the default setting. where <#> is the desired maximum number of jobs. Which value is prudent to use depends on your system and the resources allocated by Docker. Reasonable values that should run on most systems without problems are values between 2 and 8.

marcelwa:
Must have been a copy and paste issue. Commit 93df4dac516aadcf8b62e4ba5c5fd0c334111d91 should fix it.

PR #53, opened 2022-09-05
:art: Layout technology traits

marcelwa:
Added new traits of the form `has_*_technology` to check whether a layout or library utilizes the QCA, iNML, or SiDB technology.


marcelwa:
The diff coverage is actually not affected because no lines were added nor removed but reformatted.

PR #60, opened 2022-09-29
:construction_worker: Enable `ccache` for GitHub Action runners

marcelwa:
Using [`hendrikmuhs/ccache-action`](https://github.com/marketplace/actions/ccache-for-gh-actions), build times of the CI jobs are significantly sped up via caching.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           submodules: recursive
 
-      - name: ccache
+      - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1.2
         with:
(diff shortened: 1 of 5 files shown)


marcelwa:
Using MSVC, `CMake` currently does not support the `CMAKE_CXX_COMPILER_LAUNCHER` flag, thus, `ccache` (or `sccache` for that matter) is not properly called. This could only be fixed at the moment by moving away from MSVC as the generator in favor of `Unix Makefiles` or `Ninja`. In that case, however, we would lose many build configurations. I do not want to make this trade. Instead, we will accept the sub-par build performance on Windows CI for the moment hoping that a future update to `CMake` or `ccache` fixes this issue.

PR #71, opened 2022-11-15
Bug fix: FICTION_PYTHON_BINDINGS flag was ignored

wlambooy:
As a result of the flag being ignored, the build system would not be able to find the pyfiction build target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,7 @@ endif ()
 # Python Bindings
 option(FICTION_PYTHON_BINDINGS "Build fiction Python bindings" OFF)
 if (FICTION_PYTHON_BINDINGS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ endif ()
 
 # Python Bindings
 option(FICTION_PYTHON_BINDINGS "Build fiction Python bindings" OFF)
-if (FICTION_EXPERIMENTS)
+if (FICTION_PYTHON_BINDINGS)
     message(STATUS "Building fiction experiments")
```
marcelwa, line 75: This line should then probably also say "Building fiction Python bindings" instead

marcelwa (changes requested):

PR #74, opened 2022-11-24
:construction_worker: Add an IWYU CI workflow

marcelwa:
This PR adds an [IWYU](https://include-what-you-use.org/) workflow to the CI.


marcelwa:
For right now, this action does not work in the intended way. Maybe, we revisit it later.

PR #83, opened 2022-12-02
:sparkles: Multi-Path Signal Routing

marcelwa:
This PR incorporates the code from the paper *"Efficient Multi-Path Signal Routing for Field-coupled Nanotechnologies"* by M. Walter and R. Wille in NANOARCH 2022. It provides the following features:

- Reimplementations of four established path-finding algorithms ported to *fiction*'s clocked layout data structures for determining routes from a source to a target coordinate on arbitrary topologies with arbitrary clocking schemes:
  - Recursive enumeration of all possible paths
  - A* for the shortest path
  - Jump Point Search (JPS) for the shortest path (proof-of-concept)
  - k shortest paths via Yen's algorithm
- An interface for distance functors that determine (heuristic) distances between two coordinates on a layout. Pre-defined ones are:
   - Manhattan distance
   - Euclidean distance
   - A* distance
- An interface for cost functors that determine the penalty incurred from having a path be routed via a certain connection on a layout. Pre-defined ones are:
   - Unit cost
   - Random cost
- A new layout-type layer that models obstructions during pathfinding and works out-of-the-box with gate-level layouts and cell-level layouts. It can store additional obstructions in terms of blocked coordinates and connections that are being respected in the aforementioned routing algorithms
- Generation of edge intersection graphs from enumerated routing paths
- Graph coloring with a selection of SAT solvers or heuristic algorithms
- Color routing as the newly proposed multi-path routing algorithm that relies on the aforementioned implementations
- Further improvements to performance measures and memory footprints of all layout types. In particular, the `foreach_*` functions do not allocate additional memory anymore, and gate-level layout and cell-level layout have switched to `phmap::parallel_hash_map` as the main container to store coordinate association. In total, this has led to speed-ups of up to 50% on standard physical design flows.


experiments/color_routing/color_routing.cpp
```diff
@@ -0,0 +1,252 @@
...
+int main()  // NOLINT
+{
+    smt_sat_complete();
+    ortho_sat_complete();
+    ortho_mcs();
+
+    return EXIT_SUCCESS;
+}
+
+#endif  // FICTION_Z3_SOLVER
```
marcelwa, line 252: Add missing newline

PR #101, opened 2023-01-30
:bug: Fixed a bug in SAT-based graph coloring (this time for real)

marcelwa:
The bug occurred when clique information was provided for mutex cliques. The algorithm now falls back to MCS coloring.


marcelwa:
Again, many thanks to @simon1hofmann for reporting!

PR #128, opened 2023-02-21
:sparkles: Added multi-threading support for QuickSim

Drewniok:


marcelwa:
Why are you using `std::thread` instead of `std::execution_policy`?

Drewniok:
Actually, my first step was to use `std::execution_policy` since you mentioned this once. However, I was not able to make it work. As far as I can see, Apple Clang does not yet support `parallel algorithms and execution policies`.

marcelwa:
I can't say for sure and your milage may vary, but I had some luck writing a thin abstraction layer on top of `std::execution_policy` that makes the general concept available error-free on all platforms we're continuously testing on. That is, if policies are supported, they are used, if not, not. This might still not be the solution you were hoping for and it might not even work on Apple Silicon chips. That requires further testing. If you want to try it at least, please check out the [execution policy macros](https://fiction.readthedocs.io/en/placer/utils/utils.html#execution-policy-macros) on the `placer` branch.

Drewniok:
Sounds really interesting! Maybe we can keep this in mind. For now, I would suggest using the current implementation since the performance increases as expected and I have already tested/analyzed all the different benchmarks. What is your opinion?

marcelwa:
I agree with you.

marcelwa:
Could you please add a few tests for varying thread counts? After that, this PR is ready to be merged.

test/algorithms/simulation/sidb/quicksim.cpp
```diff
@@ -74,4 +74,47 @@ TEMPLATE_TEST_CASE(
...
+        lyt.assign_cell_type({1, 3, 0}, TestType::cell_type::NORMAL);
+        lyt.assign_cell_type({3, 3, 0}, TestType::cell_type::NORMAL);
+        lyt.assign_cell_type({4, 3, 0}, TestType::cell_type::NORMAL);
+
+        lyt.assign_cell_type({6, 3, 0}, TestType::cell_type::NORMAL);
+        lyt.assign_cell_type({7, 3, 0}, TestType::cell_type::NORMAL);
+
+        lyt.assign_cell_type({6, 10, 0}, TestType::cell_type::NORMAL);
+        lyt.assign_cell_type({7, 10, 0}, TestType::cell_type::NORMAL);
+
```
marcelwa, line 91: Please create one `SECTION` for each thread count number.

test/algorithms/simulation/sidb/quicksim.cpp
```diff
@@ -74,4 +74,47 @@ TEMPLATE_TEST_CASE(
...
+        quicksim_stats<TestType>     quicksimstats_100_threads{};
+        const struct quicksim_params quicksim_params_100_threads
+        {
+            params, 80, 0.7, 100
+        };
+        quicksim<TestType>(lyt, quicksim_params_100_threads, &quicksimstats_100_threads);
+        CHECK(!quicksimstats_100_threads.valid_lyts.empty());
+        CHECK(quicksimstats_100_threads.time_total.count() > 0);
+    }
 }
```
marcelwa, line 120: What happens when you specify 0 threads? Can you add a test for that?

marcelwa (changes requested):

test/algorithms/simulation/sidb/quicksim.cpp
```diff
@@ -74,4 +74,103 @@ TEMPLATE_TEST_CASE(
...
+
+        lyt.assign_cell_type({6, 3, 0}, TestType::cell_type::NORMAL);
+        lyt.assign_cell_type({7, 3, 0}, TestType::cell_type::NORMAL);
+
+        lyt.assign_cell_type({6, 10, 0}, TestType::cell_type::NORMAL);
+        lyt.assign_cell_type({7, 10, 0}, TestType::cell_type::NORMAL);
+
+        quicksim_stats<TestType>         quicksimstats{};
+        const sidb_simulation_parameters params{2, -0.30};
+        const struct quicksim_params     quicksim_params
```
marcelwa, line 94: Remove `struct` keyword. You probably don't want to define a new struct here ;) Same for the other sections

Drewniok: Thank you! Weird, because before it was required when all was in one section.

include/fiction/algorithms/simulation/sidb/quicksim.hpp
```diff
@@ -118,30 +135,60 @@ void quicksim(const Lyt& lyt, const quicksim_params& ps = quicksim_params{}, qui
...
+        // If the number of threads is initially set to zero, the simulation is run with one thread.
+        uint64_t num_threads = std::max(ps.number_threads, uint64_t{1});
+
+        // split the iterations among threads
+        const auto iter_per_thread =
+            std::max(ps.interation_steps / num_threads,
+                     uint64_t{1});  // If the number of set threads is greater than the number of iterations, the
+                                    // number of threads defines how many times QuickSim is repeated
+
+        const auto bound = static_cast<uint64_t>(std::round(0.6 * static_cast<double>(charge_lyt.num_cells())));
```
marcelwa, line 147: Can you please provide a description via comment of this line of code? Is it an upper or lower bound? What does the `0.6` multiplier do?

include/fiction/algorithms/simulation/sidb/quicksim.hpp
```diff
@@ -118,30 +135,60 @@ void quicksim(const Lyt& lyt, const quicksim_params& ps = quicksim_params{}, qui
             st.valid_lyts.push_back(charge_distribution_surface<Lyt>{charge_lyt});
         }
 
-        auto       best_energy = std::numeric_limits<double>::max();
-        const auto bound       = static_cast<uint64_t>(std::round(0.6 * static_cast<double>(charge_lyt.num_cells())));
-        for (uint64_t z = 0u; z < ps.interation_steps; z++)
+        // If the number of threads is initially set to zero, the simulation is run with one thread.
+        uint64_t num_threads = std::max(ps.number_threads, uint64_t{1});
```
marcelwa, line 139: Can be `const`

marcelwa (changes requested):

marcelwa (approved):

PR #129, opened 2023-02-21
🎨  doxygenfunction of quicksim was under the wrong header

Drewniok:


marcelwa:
This PR merely removes code. Is that correct?

Drewniok:
The doxygenfunction for quicksim was present under two headers. I removed the wrong one.

PR #134, opened 2023-03-08
:bug: Fixed accessing non-existing variables in `potential_between_sidbs`

marcelwa:
The function `potential_between_sidbs` in `charge_distribution_surface` was accessing a non-existing variable, which caused compilation to fail if this function was called. This PR fixes the behavior.


Drewniok (approved):
You are absolutely right. Thank you.

PR #177, opened 2023-04-06
:sparkles: Make ClangFormat aware of different line ending types and enforce `LF`

marcelwa:
## Description

This PR extends the ClangFormat configuration to standardize all line endings to use `LF`. It also reformats the project using ClangFormat 17.

## Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -84,5 +84,6 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard: c++17
 UseTab: Never
+LineEnding: LF
 
 ...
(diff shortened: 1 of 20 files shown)


marcelwa:
Several things here:

1) `clang-format-lint-action` only supports ClangFormat versions up to 15. I have added a version 16 feature: `LineEnding: LF`. It makes it crash... Furthermore, the action does not allow us to specify the execution parameter `--Wno-error=unknown`, which would prevent it from crashing. But that's not the only problem brought to attention by this behavior.
2) Later versions of ClangFormat (16 or 17) seem to format consecutive comments differently than version 15. This can lead to an inconsistency between our local formatter and the one in the GitHub Action.

Ideally, we find a way to use ClangFormat 17 in the CI.

marcelwa:
After raising [an issue](https://github.com/DoozyX/clang-format-lint-action/issues/53) in the respective repo, there is now support for ClangFormat 16, which we are using now. Hope, everything checks out :crossed_fingers:

PR #185, opened 2023-04-14
✨ Added Z3 toggle to the `pyfiction` build process

simon1hofmann:
## Description
Added a toggle to enable Z3 during the `setup.py` build process.

Can be invoked by inserting an environmental variable before the build process:
`z3="ON" pip install .`

Defaults to `z3="OFF"` if not specified. 

## Checklist:


setup.py
```diff
@@ -31,7 +34,7 @@ def build_extension(self, ext):
                       '-DFICTION_TEST=OFF',
                       '-DFICTION_EXPERIMENTS=OFF',
                       '-DFICTION_PYTHON_BINDINGS=ON',
-                      '-DFICTION_Z3=ON',
+                      f'-DFICTION_Z3={z3}',
```
marcelwa, line 37: I feel like the default should be OFF to be in line with the fiction flag.

marcelwa (changes requested):
If you could revert the submodule changes, that would be perfect. Many thanks!

marcelwa:
Many thanks for the edits. Can you now adjust the Python workflow accordingly to make sure it builds Z3?

Also is there a solution for the Z3-enabled test cases? Such that they are not run when Z3 is disabled? In our C++ test cases, we have implemented a solution using macros.

.github/workflows/python-bindings.yml
```diff
@@ -68,7 +69,7 @@ jobs:
 
       - name: Build pyfiction
         working-directory: ${{github.workspace}}
-        run: pip install .
+        run: z3=${{env.ENABLE_Z3}} pip install .
```
marcelwa, line 72: This looks good. Many thanks! 

Regarding the Python tests: right now, they will fail if Z3 was not enabled. Can we disable the respective ones in that case? I was thinking exposing a certain constant in the `exact` bindings and then checking for it in the Python tests for `exact`. If you know of a better way, please let me know 🙂

marcelwa (changes requested):

.github/workflows/python-bindings.yml
simon1hofmann: My idea would be to save the environment variable `z3` in an `.env` file and skip the respective tests based on that value ?

marcelwa: If that's possible, that's even better. Thank you!

setup.py
```diff
@@ -23,6 +23,13 @@ def build_extension(self, ext):
         if not extdir.endswith(os.path.sep):
             extdir += os.path.sep
 
+        # check if Z3 should be used. If no argument is given, fiction depends on Z3 by default.
```
marcelwa, line 26: This docstring seems out of date.

marcelwa (changes requested):
Looks really good! Just a tiny inconsistency: one docstring seems to have run out of date.

marcelwa (approved):
Many thanks for your contribution! LGTM.

PR #194, opened 2023-04-21
:white_check_mark: Ensured a reproducible SiDB order for test reproducibility

marcelwa:
## Description

To ensure test reproducibility on all systems, this PR enforces an SiDB order in the `write_sqd_sim_result` function.

## Checklist:

### DIFF
--- a/include/fiction/io/write_sqd_sim_result.hpp
+++ b/include/fiction/io/write_sqd_sim_result.hpp
@@ -21,7 +21,6 @@
 #include <ctime>
 #include <fstream>
 #include <functional>
-#include <memory>
 #include <ostream>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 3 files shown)


include/fiction/io/write_sqd_sim_result.hpp
```diff
@@ -205,13 +235,21 @@ class write_sqd_sim_result_impl
         std::for_each(sim_result.charge_distributions.cbegin(), sim_result.charge_distributions.cend(),
```
Drewniok, line 235: Unfortunately, there is no guarantee that the order will always be the same, since the exhaustive simulation starts with unordered maps, which are then converted to vectors. 
I just saw it because it still fails on my machine due to this problem. Thanks for your efforts!

marcelwa: Ah, I see. The surfaces themselves are also unordered. Alright, we can fix that easily enough.

Drewniok (approved):
LGTM! Thank you!

PR #219, opened 2023-05-19
:sparkles: Enable charge distribution surface printing

wlambooy:
## Description

This PR introduces a simple method for printing charge distribution surface objects to a console. An optional flag may be passed (defaulting to false) which enables colored output in the same color scheme as SiQAD.

### DIFF
--- a/include/fiction/io/print_layout.hpp
+++ b/include/fiction/io/print_layout.hpp
@@ -314,7 +314,7 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
     os << std::endl;
 }
 /**
- * Writes a simplified 2D representation of an  cell-level layout to an output stream.
+ * Writes a simplified 2D representation of a charge distribution surface to an output stream.
  *
  * @tparam Lyt Cell-level layout based in SiQAD-coordinates.
  * @param os Output stream to write into.
@@ -344,14 +344,13 @@ void print_charge_distribution_surface(std::ostream& os, const charge_distributi
 
             if (Lyt::technology::is_normal_cell(ct))
             {
-                const auto& db_locs = cds.get_all_sidb_locations_in_nm();
-
-                const auto& it = std::find(db_locs.cbegin(), db_locs.cend(),
+                const auto& it = std::find(cds.get_all_sidb_locations_in_nm().cbegin(),
+                                           cds.get_all_sidb_locations_in_nm().cend(),
                                            sidb_nm_position<Lyt>(cds.get_phys_params(), c));
 
-                if (it != db_locs.cend())
+                if (it != cds.get_all_sidb_locations_in_nm().cend())
                 {
-                    switch (cds.get_all_sidb_charges()[std::distance(db_locs.cbegin(), it)])
+                    switch (cds.get_all_sidb_charges()[std::distance(cds.get_all_sidb_locations_in_nm().cbegin(), it)])
                     {
                         case sidb_charge_state::NEGATIVE:
                         {
--- a/include/fiction/io/print_layout.hpp
+++ b/include/fiction/io/print_layout.hpp
@@ -32,6 +32,12 @@ static const std::array<fmt::text_style, 4> CLOCK_COLOR{{fmt::fg(fmt::color::bla
                                                          fmt::fg(fmt::color::black) | fmt::bg(fmt::color::light_gray),
                                                          fmt::fg(fmt::color::white) | fmt::bg(fmt::color::gray),
                                                          fmt::fg(fmt::color::white) | fmt::bg(fmt::color::dark_gray)}};
+// Escape color sequence for negatively charged SiDB colors (blue).
+static const auto SIDB_NEG_COLOR = fmt::fg(fmt::color::blue);
+// Escape color sequence for positively charged SiDB colors (red).
+static const auto SIDB_POS_COLOR = fmt::fg(fmt::color::red);
+// Escape color sequence for charge-neutral SiDB colors (white).
+static const auto SIDB_NEUT_COLOR = fmt::fg(fmt::color::red);
 
 }  // namespace detail
 
@@ -300,6 +306,82 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
 
                 os << fmt::format(color,
                                   (Lyt::technology::is_normal_cell(ct) ? "▢" : std::string(1u, static_cast<char>(ct))));
+            }
+        }
+        os << '\n';
+    }
+    // flush stream
+    os << std::endl;
+}
+/**
+ * Writes a simplified 2D representation of an  cell-level layout to an output stream.
+ *
+ * @tparam Lyt Cell-level layout based in SiQAD-coordinates.
+ * @param os Output stream to write into.
+ * @param cds The charge distribution surface to print.
+ * @param cs_color Flag to utilize color escapes for charge states.
+ */
+template <typename Lyt>
+void print_charge_distribution_surface(std::ostream& os, const charge_distribution_surface<Lyt>& cds,
+                                       const bool cs_color = false) noexcept
+{
+    // empty layout
+    if (cds.num_cells() == 0ul)
+    {
+        os << "[i] empty layout" << std::endl;
+        return;
+    }
+
+    for (decltype(cds.y()) y_pos = 0; y_pos <= cds.y(); ++y_pos)
+    {
+        for (decltype(cds.x()) x_pos = 0; x_pos <= cds.x(); ++x_pos)
+        {
+            cell<Lyt> c{x_pos, y_pos};
+
+            const fmt::text_style no_color{};
+
+            const auto ct = cds.get_cell_type(c);
+
+            if (Lyt::technology::is_normal_cell(ct))
+            {
+                const auto& db_locs = cds.get_all_sidb_locations_in_nm();
+
+                const auto& it = std::find(db_locs.cbegin(), db_locs.cend(),
+                                           sidb_nm_position<Lyt>(cds.get_phys_params(), c));
+
+                if (it != db_locs.cend())
+                {
+                    switch (cds.get_all_sidb_charges()[std::distance(db_locs.cbegin(), it)])
+                    {
+                        case sidb_charge_state::NEGATIVE:
+                        {
+                            os << fmt::format(cs_color ? detail::SIDB_NEG_COLOR : no_color, "-");
+                            continue;
+                        }
+                        case sidb_charge_state::POSITIVE:
+                        {
+                            os << fmt::format(cs_color ? detail::SIDB_POS_COLOR : no_color, "+");
+                            continue;
+                        }
+                        case sidb_charge_state::NEUTRAL:
+                        {
+                            os << fmt::format(cs_color ? detail::SIDB_NEUT_COLOR : no_color, "o");
+                            continue;
+                        }
+                        default:
+                        {
+                            os << "▢";
+                        }
+                    }
+                }
+                else
+                {
+                    os << "▢";
+                }
+            }
+            else
+            {
+                os << std::string(1u, static_cast<char>(ct));
             }
         }
         os << '\n';
--- a/include/fiction/io/print_layout.hpp
+++ b/include/fiction/io/print_layout.hpp
@@ -315,15 +315,15 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
     os << std::endl;
 }
 /**
- * Writes a simplified 2D representation of a charge distribution surface to an output stream.
+ * Writes a simplified 2D representation of a SiDB charge layout to an output stream.
  *
  * @tparam Lyt Cell-level layout based in SiQAD-coordinates.
  * @param os Output stream to write into.
- * @param cds The charge distribution surface to print.
+ * @param cds The charge distribution surface of which the charge layout is to be printed.
  * @param cs_color Flag to utilize color escapes for charge states.
  */
 template <typename Lyt>
-void print_charge_distribution_surface(std::ostream& os, const charge_distribution_surface<Lyt>& cds,
+void print_charge_layout(std::ostream& os, const charge_distribution_surface<Lyt>& cds,
                                        const bool cs_color = false) noexcept
 {
     // empty layout
--- a/include/fiction/io/print_layout.hpp
+++ b/include/fiction/io/print_layout.hpp
@@ -6,6 +6,7 @@
 #define FICTION_PRINT_LAYOUT_HPP
 
 #include "fiction/technology/charge_distribution_surface.hpp"
+#include "fiction/technology/sidb_nm_position.hpp"
 #include "fiction/traits.hpp"
 #include "fiction/types.hpp"
 
--- a/include/fiction/io/print_layout.hpp
+++ b/include/fiction/io/print_layout.hpp
@@ -5,6 +5,7 @@
 #ifndef FICTION_PRINT_LAYOUT_HPP
 #define FICTION_PRINT_LAYOUT_HPP
 
+#include "fiction/technology/charge_distribution_surface.hpp"
 #include "fiction/traits.hpp"
 #include "fiction/types.hpp"
 


include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
     // flush stream
     os << std::endl;
 }
+/**
+ * Writes a simplified 2D representation of a SiDB charge layout to an output stream.
```
marcelwa, line 320:
```suggestion
 * Writes a simplified 2D representation of an SiDB charge layout to an output stream.
```

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+ *
+ * @tparam Lyt Cell-level layout based in SiQAD-coordinates.
```
marcelwa, line 322:
```suggestion
 * @tparam Lyt Cell-level layout based on SiQAD coordinates.
```

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+ * @param os Output stream to write into.
+ * @param cds The charge distribution surface of which the charge layout is to be printed.
+ * @param cs_color Flag to utilize color escapes for charge states.
+ */
+template <typename Lyt>
+void print_charge_layout(std::ostream& os, const charge_distribution_surface<Lyt>& cds,
+                         const bool cs_color = false) noexcept
```
marcelwa, line 329: Maybe we make the coloring `true` by default?

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+        return;
+    }
+
+    for (decltype(cds.y()) y_pos = 0; y_pos <= cds.y(); ++y_pos)
+    {
+        for (decltype(cds.x()) x_pos = 0; x_pos <= cds.x(); ++x_pos)
+        {
+            cell<Lyt> c{x_pos, y_pos};
+
+            const fmt::text_style no_color{};
```
marcelwa, line 344:
```suggestion
            static const fmt::text_style no_color{};
```

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+            const auto ct = cds.get_cell_type(c);
+
+            if (Lyt::technology::is_normal_cell(ct))
+            {
+                const auto& it = std::find_if(
+                    cds.get_all_sidb_locations_in_nm().cbegin(), cds.get_all_sidb_locations_in_nm().cend(),
+                    [cur_loc = sidb_nm_position<Lyt>(cds.get_phys_params(), c)](const std::pair<double, double>& loc)
+                    {
+                        return round_to_n_decimal_places(cur_loc.first, 3) == round_to_n_decimal_places(loc.first, 3) &&
+                               round_to_n_decimal_places(cur_loc.second, 3) == round_to_n_decimal_places(loc.second, 3);
```
marcelwa, line 355: I must admit that I have no idea what's happening here :sweat_smile: Why do you need the nanometer positions? The SiQAD coordinates should be enough to provide grid location information.

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+                        }
+                        case sidb_charge_state::NEUTRAL:
+                        {
+                            os << fmt::format(cs_color ? detail::SIDB_NEUT_COLOR : no_color, "o");
+                            continue;
+                        }
+                        default:
+                        {
+                            os << "▢";
+                        }
```
marcelwa, line 381: Which state is left for this line to be triggered?

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +316,86 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+                    }
+                }
+                else
+                {
+                    os << "▢";
+                }
```
marcelwa, line 387: What's happening here?

marcelwa (changes requested):

include/fiction/io/print_layout.hpp
wlambooy: Fixed this in the next commit, though I'm waiting for the tests this time.

include/fiction/io/print_layout.hpp
wlambooy: Fixed this in the next commit, though I'm waiting for the tests this time.

include/fiction/io/print_layout.hpp
wlambooy: Fixed this in the next commit, though I'm waiting for the tests this time.

include/fiction/io/print_layout.hpp
wlambooy: I reworked this in the last commit which now avoids any lookup that can trigger an exception. The problem lied in accessing the charge state based on SiQAD coordinates, since you need the index associated with the DB at a given SiQAD coordinate. I reworked it to sort a vector of indices on the order of how they would appear when printed, after this it is easy to find the index associated with the current normal cell; we simply increment the counter that indexes our indices vector each time we find a normal cell. The next commit will have a few comments to clarify this.

include/fiction/io/print_layout.hpp
wlambooy: Outdated, I reworked this so this exception state cannot occur anymore.

include/fiction/io/print_layout.hpp
wlambooy: This is for the NONE charge state. Must be included to cover all SiDB charge states. In the current commit I updated the symbols to the following:

NEGATIVE -> ●
POSITIVE -> ⨁
NEUTRAL -> ◯
NONE -> ◌

The last one is a dotted circle, though the rendering here is different than seen in terminals (that I tried at least).

In the terminals I tried where ▢ was supported, these symbols were also supported and take up a single standard unit.

wlambooy:
image
Updating negatively charged SiDB colors from blue (too dark) to cyan. This is also more in accordance to SiQAD's color scheme.

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +317,110 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+                  }
+                  return p1.second.first < p2.second.first;
+              });
+
+    // obtain fiction coordinates of the minimum and maximum normal cell
+    auto min_x = static_cast<int>(std::floor(min_x_d / cds.get_phys_params().lat_a * 10));
+    auto max_x = static_cast<int>(std::floor(max_x_d / cds.get_phys_params().lat_a * 10));
+
+    auto min_y = static_cast<int>(std::floor(sorted_locs.front().second.second / cds.get_phys_params().lat_b * 10));
+    auto max_y = static_cast<int>(std::floor(sorted_locs.back().second.second / cds.get_phys_params().lat_b * 10));
```
marcelwa, line 368: Could the utility function [`siqad::to_fiction_coord`](https://fiction.readthedocs.io/en/latest/layouts/coordinates.html#_CPPv4I0EN7fiction5siqad16to_fiction_coordE14CoordinateTypeRK7coord_t) be helpful here?

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +317,110 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+    auto min_x_d = std::numeric_limits<double>::max();
+    auto max_x_d = std::numeric_limits<double>::min();
+
+    for (uint64_t i = 0; i < cds.num_cells(); ++i)
+    {
+        min_x_d = std::min(min_x_d, cds.get_all_sidb_locations_in_nm()[i].first);
+        max_x_d = std::max(max_x_d, cds.get_all_sidb_locations_in_nm()[i].first);
+
+        sorted_locs.push_back(std::make_pair(i, cds.get_all_sidb_locations_in_nm()[i]));
+    };
```
marcelwa, line 350: I think you could use `cds.foreach_cell` here and then simply add each cell to a vector.

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +317,110 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+    // sort a vector containing indices on the order of how they are printed
+    std::sort(sorted_locs.begin(), sorted_locs.end(),
+              [](const auto& p1, const auto& p2)
+              {
+                  if (p1.second.second != p2.second.second)
+                  {
+                      return p1.second.second < p2.second.second;
+                  }
+                  return p1.second.first < p2.second.first;
+              });
```
marcelwa, line 361: On all coordinate types, the `operator<` is implemented, which means you can sort them without the need for an extra lambda.

include/fiction/io/print_layout.hpp
```diff
@@ -307,6 +317,110 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+    auto min_x = static_cast<int>(std::floor(min_x_d / cds.get_phys_params().lat_a * 10));
+    auto max_x = static_cast<int>(std::floor(max_x_d / cds.get_phys_params().lat_a * 10));
+
+    auto min_y = static_cast<int>(std::floor(sorted_locs.front().second.second / cds.get_phys_params().lat_b * 10));
+    auto max_y = static_cast<int>(std::floor(sorted_locs.back().second.second / cds.get_phys_params().lat_b * 10));
+
+    // initialize the count which indexes the sorted vector containing the indices associated with the cells
+    uint64_t count = 0;
+
+    static const fmt::text_style no_color{};
```
marcelwa, line 373: Maybe move this up to the other constants as well?

marcelwa (changes requested):

marcelwa:
@wlambooy the output looks really cool! 🤩 Many thanks already for the great effort!

include/fiction/io/print_layout.hpp
wlambooy: I implemented it in a new commit

include/fiction/io/print_layout.hpp
wlambooy: I implemented it in a new commit

include/fiction/io/print_layout.hpp
wlambooy: I implemented it in a new commit

include/fiction/io/print_layout.hpp
wlambooy: By using foreach_cell in a new commit, I was able to circumvent this conversion step.

test/io/print_layout.cpp
```diff
@@ -194,3 +197,114 @@ TEST_CASE("Print wire crossing cell-level layout", "[print-cell-level-layout]")
...
+    cl.assign_charge_state_by_cell_index(8, sidb_charge_state::POSITIVE, false);
+    cl.assign_charge_state_by_cell_index(9, sidb_charge_state::NEGATIVE, false);
+    cl.assign_charge_state_by_cell_index(10, sidb_charge_state::NEUTRAL, false);
+    cl.assign_charge_state_by_cell_index(11, sidb_charge_state::NEGATIVE, false);
+    cl.assign_charge_state_by_cell_index(12, sidb_charge_state::NONE, false);
+    cl.assign_charge_state_by_cell_index(13, sidb_charge_state::NEGATIVE, false);
+    cl.assign_charge_state_by_cell_index(14, sidb_charge_state::NEUTRAL, false);
+    cl.assign_charge_state_by_cell_index(15, sidb_charge_state::NEGATIVE, false);
+    cl.assign_charge_state_by_cell_index(16, sidb_charge_state::POSITIVE, false);
+    cl.assign_charge_state_by_cell_index(17, sidb_charge_state::NEGATIVE, false);
```
marcelwa, line 243: Same here. You could just use `assign_charge_state`

marcelwa (changes requested):
Many thanks for the iteration. The code is so much better already!

include/fiction/io/print_layout.hpp
```diff
@@ -5,13 +5,15 @@
...
 #include "fiction/traits.hpp"
 #include "fiction/types.hpp"
 
 #include <fmt/color.h>
 #include <fmt/format.h>
 
 #include <array>
+#include <cmath>
 #include <string>
 #include <vector>
```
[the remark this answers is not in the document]
wlambooy: Newest commit implements this change

include/fiction/io/print_layout.hpp
```diff
@@ -304,6 +317,107 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+    auto max_x = std::numeric_limits<decltype(cds.x())>::min();
+
+    std::vector<std::pair<uint64_t, coordinate<Lyt>>> sorted_locs;
+
+    auto ix = 0;
+
+    cds.foreach_cell(
+        [&cds, &min_x, &max_x, &sorted_locs, &ix](const cell<Lyt>& c)
+        {
+            if (Lyt::technology::is_empty_cell(cds.get_cell_type(c)))
```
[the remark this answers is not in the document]
wlambooy: Newest commit implements this change

include/fiction/io/print_layout.hpp
```diff
@@ -304,6 +317,107 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+    if (cds.num_cells() == 0ul)
+    {
+        os << "[i] empty layout" << std::endl;
+        return;
+    }
+
+    auto min_x = std::numeric_limits<decltype(cds.x())>::max();
+    auto max_x = std::numeric_limits<decltype(cds.x())>::min();
+
+    std::vector<std::pair<uint64_t, coordinate<Lyt>>> sorted_locs;
```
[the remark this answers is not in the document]
wlambooy: `sorted_locs` is removed in the newest commit

include/fiction/io/print_layout.hpp
```diff
@@ -304,6 +317,107 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+    uint64_t count = 0;
+
+    for (decltype(cds.y()) y_pos = min.y; y_pos <= max.y; ++y_pos)
+    {
+        // loop over the different rows of a dimer pair
+        for (uint8_t r = 0; r <= 1; ++r)
+        {
+            for (decltype(cds.x()) x_pos = min.x; x_pos <= max.x; ++x_pos)
+            {
+                if (Lyt::technology::is_empty_cell(cds.get_cell_type({x_pos, y_pos, r})))
```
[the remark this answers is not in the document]
wlambooy: Newest commit implements this change

test/io/print_layout.cpp
wlambooy: Newest commit implements this change

include/fiction/io/print_layout.hpp
```diff
@@ -304,6 +317,107 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+ * @param cds The charge distribution surface of which the charge layout is to be printed.
+ * @param cs_color Flag to utilize color escapes for charge states.
+ * @param draw_lattice Flag to enable lattice background drawing.
+ */
+template <typename Lyt>
+void print_charge_layout(std::ostream& os, const charge_distribution_surface<Lyt>& cds, const bool cs_color = true,
+                         const bool draw_lattice = true) noexcept
+{
+    // empty layout
+    if (cds.num_cells() == 0ul)
```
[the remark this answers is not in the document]
wlambooy: Newest commit implements this change

include/fiction/io/print_layout.hpp
```diff
@@ -304,6 +317,107 @@ void print_cell_level_layout(std::ostream& os, const Lyt& layout, const bool io_
...
+            {
+                if (Lyt::technology::is_empty_cell(cds.get_cell_type({x_pos, y_pos, r})))
+                {
+                    os << (draw_lattice ? fmt::format(cs_color ? detail::SIDB_LAT_COLOR : detail::NO_COLOR, " · ") :
+                                          "  ");
+                    continue;
+                }
+
+                // switch over the charge state of the SiDB index associated with the current cell, and update count
+                switch (cds.get_charge_state_by_index(sorted_locs[count++].first))
```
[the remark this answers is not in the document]
wlambooy: Newest commit implements this change

wlambooy:
One check failed due to a failing test in `test_write_sqd_layout`. The difference between the read and the written sqd file  for which the comparison fails is a single second difference in their respective time stamps. Makes sense that it is possible that this can occur. 

Seeing that it has nothing to do with this PR, could the checks be run again? (if at all possible)

include/fiction/io/print_layout.hpp
```diff
@@ -32,6 +39,16 @@ static const std::array<fmt::text_style, 4> CLOCK_COLOR{{fmt::fg(fmt::color::bla
                                                          fmt::fg(fmt::color::black) | fmt::bg(fmt::color::light_gray),
                                                          fmt::fg(fmt::color::white) | fmt::bg(fmt::color::gray),
                                                          fmt::fg(fmt::color::white) | fmt::bg(fmt::color::dark_gray)}};
+// Escape color sequence for negatively charged SiDB colors (blue).
```
marcelwa, line 42: I think this comment has gone out of sync.

include/fiction/io/print_layout.hpp
```diff
@@ -32,6 +39,16 @@ static const std::array<fmt::text_style, 4> CLOCK_COLOR{{fmt::fg(fmt::color::bla
...
+static const auto SIDB_NEG_COLOR = fmt::fg(fmt::color::cyan);
+// Escape color sequence for positively charged SiDB colors (red).
+static const auto SIDB_POS_COLOR = fmt::fg(fmt::color::red);
+// Escape color sequence for charge-neutral SiDB colors (white).
+static const auto SIDB_NEUT_COLOR = fmt::fg(fmt::color::white);
+// Escape color sequence for lattice background colors (grey).
+static const auto SIDB_LAT_COLOR = fmt::fg(fmt::color::gray);
+// Empty escape color sequence
+static constexpr auto NO_COLOR = fmt::text_style{};
```
marcelwa, line 51: Seeing that `NO_COLOR` could be `constexpr`, I'm wondering whether the other definitions can be `constexpr` as well. Also, if they are defined in the global scope, we might want to make them `inline constexpr` instead of `static constexpr`.

include/fiction/io/print_layout.hpp
```diff
@@ -5,14 +5,21 @@
...
 #include <fmt/format.h>
 
+#include <algorithm>
 #include <array>
+#include <cmath>
+#include <cstdint>
+#include <iostream>
+#include <limits>
 #include <string>
+#include <utility>
```
marcelwa, line 22: Now that you removed the usage of `std::pair`, you do not need the `<utility>` header anymore.

marcelwa (changes requested):
Looks amazing! Many thanks for the quick changes. Now we're down to cosmetic adjustments. 

Also, could you add the `print_charge_layout` function to the RST documentation?

include/fiction/io/print_layout.hpp
wlambooy: Fixed in the latest commit

include/fiction/io/print_layout.hpp
wlambooy: `NO_COLOR` is `inline constexpr` in the latest commit; the others had to stay this way since `fmt::fg` cannot be in a `constexpr` it seems

marcelwa (approved):

PR #233, opened 2023-06-07
✨ Add function to generate a random coordinate.

Drewniok:
## Description

This PR adds a function to create a random cell between two given cells.

## Checklist:


include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +351,56 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
 
     return lyt_new;
 }
+/**
+ * Generates a random cell between two given cells.
```
marcelwa, line 355: What does "between" really mean here? Could you specify the functionality more elaborately in the docstring, please?

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +351,56 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+ *
+ * @tparam Lyt Cell-level layout type.
```
marcelwa, line 357: Why does it have to be cell-level? Couldn't we use coordinates instead to make it more general?

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +351,56 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+ * @param cell1 First given cell.
+ * @return cell2 Second given cell.
```
marcelwa, line 359: "First" and "Second" might not be very descriptive attributes of the passed entities.

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +351,56 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+ */
+template <typename Lyt>
+typename Lyt::cell random_cell(const typename Lyt::cell& cell1, const typename Lyt::cell& cell2)
```
marcelwa, line 362: You should be able to use `coordinate<Lyt>` syntax here

marcelwa, line 362:
```suggestion
typename Lyt::cell random_cell(const typename Lyt::cell& cell1, const typename Lyt::cell& cell2) noexcept
```

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +351,56 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+    if (cell1 <= cell2)
+    {
+        left_cell  = cell1;
+        right_cell = cell2;
+    }
+    else
+    {
+        left_cell  = cell2;
+        right_cell = cell1;
+    }
```
marcelwa, line 376: Rather pass the variables by copy and use `std::swap` if the coordinates are in the "wrong" order

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +351,56 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+
+    static std::mt19937_64 generator(std::random_device{}());
```
marcelwa, line 378: Move `static` variables to the top of the function

marcelwa (changes requested):

include/fiction/utils/layout_utils.hpp
Drewniok: Thank you! That's convenient.

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +352,54 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+ * necessary).
+ * return Randomly generated coordinate.
+ */
+template <typename CoordinateType>
+CoordinateType random_coordinate(const CoordinateType& coordinate1, const CoordinateType& coordinate2) noexcept
+{
+    static std::mt19937_64 generator(std::random_device{}());
+
+    CoordinateType left_coordinate  = coordinate1;
+    CoordinateType right_coordinate = coordinate2;
```
marcelwa, line 371: As mentioned in the last review, if you pass the parameters by copy, you don't have to allocate extra variables.

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +352,54 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
 
     return lyt_new;
 }
+/**
+ * Generates a random coordinate within the region spanned by two given coordinates. The two given coordinates form the
+ * lower left corner and the upper right corner of the spanned region.
+ *
+ * @OffsetCoordinateType The coordinate implementation to be used.
+ * @param coordinate1 Coordinate of the lower left corner.
+ * @param coordinate2 Coordinate of the upper right corner (coordinate order is not important, automatically swapped if
```
marcelwa, line 361: In fiction, coordinates usually increase in value from top left to bottom right. Is there a reason that coordinates in this function are seemingly handled differently?

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +352,54 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+    }
+    else if constexpr (std::is_same_v<CoordinateType, siqad::coord_t>)
+    {
+        std::uniform_int_distribution<> dist_x(left_coordinate.x, right_coordinate.x);
+        std::uniform_int_distribution<> dist_y(left_coordinate.y, right_coordinate.y);
+        std::uniform_int_distribution<> dist_z(left_coordinate.z, right_coordinate.z);
+
+        return {dist_x(generator), dist_y(generator), dist_z(generator)};
+    }
+}
```
marcelwa, line 402: This function contains a branch that violates its return type. If `CoordinateType` is neither `offset::ucoord_t` nor `cube::coord_t` nor `siqad::coord_t`, this function will cause a `SEGFAULT` due to the missing `return` statement here.

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +352,54 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+ *
+ * @OffsetCoordinateType The coordinate implementation to be used.
+ * @param coordinate1 Coordinate of the lower left corner.
+ * @param coordinate2 Coordinate of the upper right corner (coordinate order is not important, automatically swapped if
+ * necessary).
+ * return Randomly generated coordinate.
+ */
+template <typename CoordinateType>
+CoordinateType random_coordinate(const CoordinateType& coordinate1, const CoordinateType& coordinate2) noexcept
+{
```
marcelwa, line 367: Shouldn't we `static_assert` something here?

test/utils/layout_utils.cpp
```diff
@@ -232,3 +232,109 @@ TEMPLATE_TEST_CASE("Convert SiQAD layout to cube::coord_t coordinate layout", "[
         CHECK(lyt_transformed.get_cell_name({5, 6}) == "output cell");
     }
 }
+
+TEMPLATE_TEST_CASE("Generate random offset::ucoord_t coordinate", "[layout-utils]", (offset::ucoord_t))
```
marcelwa, line 236:
```suggestion
TEMPLATE_TEST_CASE("Generate random coordinate", "[layout-utils]", (offset::ucoord_t, cube::coord_t, siqad::coord_t))
```

marcelwa (changes requested):

include/fiction/utils/layout_utils.hpp
Drewniok: Sorry! Now, I got it.

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,7 +352,54 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+        return {dist_x(generator), dist_y(generator), dist_z(generator)};
+    }
+    else if constexpr (std::is_same_v<CoordinateType, siqad::coord_t>)
+    {
+        std::uniform_int_distribution<> dist_x(left_coordinate.x, right_coordinate.x);
+        std::uniform_int_distribution<> dist_y(left_coordinate.y, right_coordinate.y);
+        std::uniform_int_distribution<> dist_z(left_coordinate.z, right_coordinate.z);
+
+        return {dist_x(generator), dist_y(generator), dist_z(generator)};
+    }
```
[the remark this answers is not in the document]
Drewniok: Good point! They just got the same after I removed the integer type in std::uniform_int_distribution<> (I was having trouble passing all the checks in the workflow). Should be fine since all the boundaries are determined by the coordinates themselves, right?

include/fiction/utils/layout_utils.hpp
Drewniok: I tried it but then, swap does not work anymore.

marcelwa: It will if you pass the parameters by copy and remove `const`.

include/fiction/utils/layout_utils.hpp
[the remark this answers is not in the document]
marcelwa: The coordinates have intrinsic min/max values, indeed.

include/fiction/utils/layout_utils.hpp
Drewniok: True! Thanks!

test/utils/layout_utils.cpp
marcelwa: If that is the case, then this `TEST_CASE` does not need to be a templated one, right? But then also, how are the other coordinates tested?

include/fiction/utils/layout_utils.hpp
```diff
@@ -350,6 +353,36 @@ Lyt convert_to_fiction_coordinates(const sidb_cell_clk_lyt_siqad& lyt) noexcept
...
+    static std::mt19937_64 generator(std::random_device{}());
+
+    if (coordinate1 > coordinate2)
+    {
+        std::swap(coordinate1, coordinate2);
+    }
+
+    std::uniform_int_distribution<> dist_x(coordinate1.x, coordinate2.x);
+    std::uniform_int_distribution<> dist_y(coordinate1.y, coordinate2.y);
+    std::uniform_int_distribution<> dist_z(coordinate1.z, coordinate2.z);
```
marcelwa, line 382: I think this might break for SiQAD coordinates. If `coordinate2` is not located on a `z = 1` dimer atom, this function will always produce lattice coordinates on the `z = 0` dimer atoms. Is this intended behavior?

marcelwa (changes requested):

PR #255, opened 2023-07-21
:sparkles: SiDB Operational Domain Computation

marcelwa:
## Description

🚧 This PR is work-in-progress 🚧

There will be SiDB operational domain computation algorithms eventually.

## Checklist:

### DIFF
--- a/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
+++ b/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
@@ -18,22 +18,52 @@
 namespace fiction
 {
 
+/**
+ * A Binary-dot Logic (BDL) pair is a pair of SiDBs that are close to each other and, thus, most likely share a charge.
+ *
+ * @tparam Lyt SiDB cell-level layout type.
+ */
 template <typename Lyt>
 struct bdl_pair
 {
+    /**
+     * The type of the SiDBs in the pair. BDL SiDBs must be of the same type. They can either be normal, input, or
+     * output SiDBs.
+     */
     const sidb_technology::cell_type type{};
-
-    const cell<Lyt> top_sidb{};
-    const cell<Lyt> bottom_sidb{};
-
+    /**
+     * The top SiDB of the pair. Top and bottom are defined relative to each other via the `operator<` overload.
+     */
+    const cell<Lyt> top{};
+    /**
+     * The bottom SiDB of the pair. Top and bottom are defined relative to each other via the `operator<` overload.
+     */
+    const cell<Lyt> bottom{};
+    /**
+     * Standard constructor for empty BDL pairs.
+     */
     bdl_pair() = default;
-    bdl_pair(const sidb_technology::cell_type t, const cell<Lyt> top, const cell<Lyt> bottom) noexcept :
+    /**
+     * Constructor for BDL pairs.
+     *
+     * @param t Type of the SiDBs in the pair.
+     * @param upper The top SiDB of the pair.
+     * @param lower The bottom SiDB of the pair.
+     */
+    bdl_pair(const sidb_technology::cell_type t, const cell<Lyt> upper, const cell<Lyt> lower) noexcept :
             type{t},
-            top_sidb{top},
-            bottom_sidb{bottom}
-    {}
+            top{upper},
+            bottom{lower}
+    {
+        static_assert(is_cell_level_layout_v<Lyt>, "Lyt is not a cell-level layout");
+        static_assert(has_sidb_technology_v<Lyt>, "Lyt is not an SiDB layout");
+        static_assert(has_siqad_coord_v<Lyt>, "Lyt is not based on SiQAD coordinates");
+    }
 };
 
+/**
+ * Parameters for the BDL pair detection algorithms.
+ */
 struct detect_bdl_pairs_params
 {
     /**
@@ -42,6 +72,17 @@ struct detect_bdl_pairs_params
     units::length::nanometer_t threshold{2_nm};
 };
 
+/**
+ * This algorithm detects input or output BDL pairs in an SiDB layout. It does so by first collecting all input and
+ * output dots and then uniquely pairing them up based on their distance. A threshold can be defined (default = 2 nm),
+ * after which SiDBs are no longer considered a pair. The distance between two dots is computed using the
+ * `sidb_nanometer_distance` function. The algorithm returns a vector of I/O BDL pairs.
+ *
+ * @tparam Lyt SiDB cell-level layout type.
+ * @param lyt The layout to detect I/O BDL pairs in.
+ * @param params Parameters for the BDL pair detection algorithm.
+ * @return A vector of I/O BDL pairs.
+ */
 template <typename Lyt>
 std::vector<bdl_pair<Lyt>> detect_io_bdl_pairs(const Lyt& lyt, const detect_bdl_pairs_params params = {}) noexcept
 {
--- a/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
+++ b/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
@@ -1,0 +1,151 @@
+//
+// Created by marcel on 21.07.23.
+//
+
+#ifndef FICTION_DETECT_BDL_PAIRS_HPP
+#define FICTION_DETECT_BDL_PAIRS_HPP
+
+#include "fiction/algorithms/path_finding/distance.hpp"
+#include "fiction/technology/cell_technologies.hpp"
+#include "fiction/traits.hpp"
+
+#include <units.h>
+
+#include <algorithm>
+#include <unordered_set>
+#include <vector>
+
+namespace fiction
+{
+
+template <typename Lyt>
+struct bdl_pair
+{
+    const sidb_technology::cell_type type{};
+
+    const cell<Lyt> top_sidb{};
+    const cell<Lyt> bottom_sidb{};
+
+    bdl_pair() = default;
+    bdl_pair(const sidb_technology::cell_type t, const cell<Lyt> top, const cell<Lyt> bottom) noexcept :
+            type{t},
+            top_sidb{top},
+            bottom_sidb{bottom}
+    {}
+};
+
+struct detect_bdl_pairs_params
+{
+    /**
+     * The maximum distance between two dots to be considered a BDL pair.
+     */
+    units::length::nanometer_t threshold{2_nm};
+};
+
+template <typename Lyt>
+std::vector<bdl_pair<Lyt>> detect_io_bdl_pairs(const Lyt& lyt, const detect_bdl_pairs_params params = {}) noexcept
+{
+    std::vector<bdl_pair<Lyt>> bdl_pairs{};
+    bdl_pairs.reserve(lyt.num_pis() / 2 + lyt.num_pos() / 2);
+
+    const auto pair_up_dots = [&lyt, &params, &bdl_pairs](const std::vector<cell<Lyt>>&    io_dots,
+                                                          const sidb_technology::cell_type type) noexcept
+    {
+        struct pairwise_dot_distance
+        {
+            cell<Lyt> sidb1{};
+            cell<Lyt> sidb2{};
+
+            units::length::nanometer_t distance{};
+
+            pairwise_dot_distance() = default;
+            pairwise_dot_distance(const cell<Lyt> s1, const cell<Lyt> s2, const units::length::nanometer_t d) noexcept :
+                    sidb1{s1},
+                    sidb2{s2},
+                    distance{d}
+            {}
+        };
+
+        const auto compute_pairwise_dot_distances = [&lyt](const std::vector<cell<Lyt>>& dots) noexcept
+        {
+            std::vector<pairwise_dot_distance> pairwise_distances{};
+            pairwise_distances.reserve(dots.size() * (dots.size() - 1) / 2);
+
+            for (auto i = 0u; i < dots.size(); ++i)
+            {
+                for (auto j = i + 1; j < dots.size(); ++j)
+                {
+                    pairwise_distances.emplace_back(dots[i], dots[j], sidb_nanometer_distance(lyt, dots[i], dots[j]));
+                }
+            }
+
+            return pairwise_distances;
+        };
+
+        const auto dot_distance_comparator = [](const auto& lhs, const auto& rhs) noexcept
+        { return lhs.distance < rhs.distance; };
+
+        // compute pairwise distances
+        auto input_pairwise_distances = compute_pairwise_dot_distances(io_dots);
+        // sort pairwise distances
+        std::sort(input_pairwise_distances.begin(), input_pairwise_distances.end(), dot_distance_comparator);
+        // pair unique dots with the smallest distance
+        std::unordered_set<cell<Lyt>> paired_dots{};
+        paired_dots.reserve(io_dots.size());
+        const auto already_paired_up = [&paired_dots](const auto& dot) noexcept
+        { return paired_dots.find(dot) != paired_dots.cend(); };
+
+        for (auto& potential_bdl_pair : input_pairwise_distances)
+        {
+            // if the distance is larger than the threshold, we can break the loop because the remaining distances
+            // must be larger as well due to the prior sorting; this prevents unlikely pairings and helps performance
+            if (potential_bdl_pair.distance > params.threshold)
+            {
+                break;
+            }
+
+            // if either dot has already been matched, skip
+            if (already_paired_up(potential_bdl_pair.sidb1) || already_paired_up(potential_bdl_pair.sidb2))
+            {
+                continue;
+            }
+
+            // a BDL pair has been detected (swap SiDBs if necessary)
+            if (potential_bdl_pair.sidb1 > potential_bdl_pair.sidb2)
+            {
+                bdl_pairs.emplace_back(type, potential_bdl_pair.sidb2, potential_bdl_pair.sidb1);
+            }
+            else
+            {
+                bdl_pairs.emplace_back(type, potential_bdl_pair.sidb1, potential_bdl_pair.sidb2);
+            }
+
+            // mark the dots as paired
+            paired_dots.insert(potential_bdl_pair.sidb1);
+            paired_dots.insert(potential_bdl_pair.sidb2);
+        }
+    };
+
+    // collect all input dots
+    std::vector<cell<Lyt>> input_dots{};
+    input_dots.reserve(lyt.num_pis());
+    lyt.foreach_pi([&input_dots](const auto& pi) { input_dots.push_back(pi); });
+
+    // pair up input dots
+    pair_up_dots(input_dots, sidb_technology::cell_type::INPUT);
+
+    // collect all output dots
+    std::vector<cell<Lyt>> output_dots{};
+    output_dots.reserve(lyt.num_pos());
+    lyt.foreach_po([&output_dots](const auto& po) { output_dots.push_back(po); });
+
+    // pair up output dots
+    pair_up_dots(output_dots, sidb_technology::cell_type::OUTPUT);
+
+    // return the detected I/O BDL pairs
+    return bdl_pairs;
+}
+
+};      // namespace fiction
+
+#endif  // FICTION_DETECT_BDL_PAIRS_HPP
--- a/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
+++ b/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
@@ -98,9 +98,8 @@ std::vector<bdl_pair<Lyt>> detect_io_bdl_pairs(const Lyt& lyt, const detect_bdl_
      * then sorting them. The smallest distances are then used to pair up the dots. The function takes a vector of dots
      * and a cell type as input. It fills the vector of BDL pairs declared above.
      */
-    const auto pair_up_dots = [&lyt, &params,
-                               &bdl_pairs](const std::vector<cell<Lyt>>&    io_dots,
-                                           const sidb_technology::cell_type type) noexcept -> void
+    const auto pair_up_dots = [&lyt, &params, &bdl_pairs](const std::vector<cell<Lyt>>&    io_dots,
+                                                          const sidb_technology::cell_type type) noexcept -> void
     {
         /**
          * Container for pairwise dot distances used in the pairing algorithm.
--- a/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
+++ b/include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
@@ -32,13 +32,13 @@ struct bdl_pair
      */
     const sidb_technology::cell_type type{};
     /**
-     * The top SiDB of the pair. Top and bottom are defined relative to each other via the `operator<` overload.
+     * The upper SiDB of the pair. Upper and lower are defined relative to each other via the `operator<` overload.
      */
-    const cell<Lyt> top{};
+    const cell<Lyt> upper{};
     /**
-     * The bottom SiDB of the pair. Top and bottom are defined relative to each other via the `operator<` overload.
+     * The lower SiDB of the pair. Upper and lower are defined relative to each other via the `operator<` overload.
      */
-    const cell<Lyt> bottom{};
+    const cell<Lyt> lower{};
     /**
      * Standard constructor for empty BDL pairs.
      */
@@ -47,13 +47,13 @@ struct bdl_pair
      * Constructor for BDL pairs.
      *
      * @param t Type of the SiDBs in the pair.
-     * @param upper The top SiDB of the pair.
-     * @param lower The bottom SiDB of the pair.
+     * @param u The upper SiDB of the pair.
+     * @param l The lower SiDB of the pair.
      */
-    bdl_pair(const sidb_technology::cell_type t, const cell<Lyt> upper, const cell<Lyt> lower) noexcept :
+    bdl_pair(const sidb_technology::cell_type t, const cell<Lyt> u, const cell<Lyt> l) noexcept :
             type{t},
-            top{upper},
-            bottom{lower}
+            upper{u},
+            lower{l}
     {
         static_assert(is_cell_level_layout_v<Lyt>, "Lyt is not a cell-level layout");
         static_assert(has_sidb_technology_v<Lyt>, "Lyt is not an SiDB layout");
@@ -86,28 +86,61 @@ struct detect_bdl_pairs_params
 template <typename Lyt>
 std::vector<bdl_pair<Lyt>> detect_io_bdl_pairs(const Lyt& lyt, const detect_bdl_pairs_params params = {}) noexcept
 {
+    static_assert(is_cell_level_layout_v<Lyt>, "Lyt is not a cell-level layout");
+    static_assert(has_sidb_technology_v<Lyt>, "Lyt is not an SiDB layout");
+    static_assert(has_siqad_coord_v<Lyt>, "Lyt is not based on SiQAD coordinates");
+
     std::vector<bdl_pair<Lyt>> bdl_pairs{};
     bdl_pairs.reserve(lyt.num_pis() / 2 + lyt.num_pos() / 2);
 
-    const auto pair_up_dots = [&lyt, &params, &bdl_pairs](const std::vector<cell<Lyt>>&    io_dots,
-                                                          const sidb_technology::cell_type type) noexcept
+    /**
+     * Pairs up dots based on their distance. It does so by first computing the pairwise distances between all dots and
+     * then sorting them. The smallest distances are then used to pair up the dots. The function takes a vector of dots
+     * and a cell type as input. It fills the vector of BDL pairs declared above.
+     */
+    const auto pair_up_dots = [&lyt, &params,
+                               &bdl_pairs](const std::vector<cell<Lyt>>&    io_dots,
+                                           const sidb_technology::cell_type type) noexcept -> void
     {
+        /**
+         * Container for pairwise dot distances used in the pairing algorithm.
+         */
         struct pairwise_dot_distance
         {
+            /**
+             * First dot.
+             */
             cell<Lyt> sidb1{};
+            /**
+             * Second dot.
+             */
             cell<Lyt> sidb2{};
-
+            /**
+             * Distance between the two dots.
+             */
             units::length::nanometer_t distance{};
-
+            /**
+             * Standard constructor for empty pairwise dot distances.
+             */
             pairwise_dot_distance() = default;
+            /**
+             * Constructor for pairwise dot distances.
+             *
+             * @param s1 The first dot.
+             * @param s2 The second dot.
+             * @param d The distance between the two dots.
+             */
             pairwise_dot_distance(const cell<Lyt> s1, const cell<Lyt> s2, const units::length::nanometer_t d) noexcept :
                     sidb1{s1},
                     sidb2{s2},
                     distance{d}
             {}
         };
-
-        const auto compute_pairwise_dot_distances = [&lyt](const std::vector<cell<Lyt>>& dots) noexcept
+        /**
+         * Computes the pairwise distances between all dots in the input vector.
+         */
+        const auto compute_pairwise_dot_distances =
+            [&lyt](const std::vector<cell<Lyt>>& dots) noexcept -> std::vector<pairwise_dot_distance>
         {
             std::vector<pairwise_dot_distance> pairwise_distances{};
             pairwise_distances.reserve(dots.size() * (dots.size() - 1) / 2);
@@ -122,8 +155,10 @@ std::vector<bdl_pair<Lyt>> detect_io_bdl_pairs(const Lyt& lyt, const detect_bdl_
 
             return pairwise_distances;
         };
-
-        const auto dot_distance_comparator = [](const auto& lhs, const auto& rhs) noexcept
+        /**
+         * Comparator for pairwise dot distances. Used in the sorting algorithm.
+         */
+        const auto dot_distance_comparator = [](const auto& lhs, const auto& rhs) noexcept -> bool
         { return lhs.distance < rhs.distance; };
 
         // compute pairwise distances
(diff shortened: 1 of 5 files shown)


marcelwa:
@Drewniok as you can see, this is WIP code, but I thought I assign you for an early review to keep the workload small.

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
```diff
@@ -0,0 +1,123 @@
+//
+// Created by marcel on 21.07.23.
+//
+
+#include <catch2/catch_template_test_macros.hpp>
+
+#include <fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp>
+#include <fiction/io/print_layout.hpp>
```
Drewniok, line 8: I think this header is redundant.

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
```diff
@@ -0,0 +1,123 @@
...
+    CHECK(input_pair1.bottom == cell<layout>{18, 1, 1});
+
+    CHECK(input_pair2.type == layout::cell_type::INPUT);
+    CHECK(input_pair2.top == cell<layout>{0, 0, 1});
+    CHECK(input_pair2.bottom == cell<layout>{2, 1, 1});
+
+    CHECK(output_pair.type == layout::cell_type::OUTPUT);
+    CHECK(output_pair.top == cell<layout>{10, 6, 0});
+    CHECK(output_pair.bottom == cell<layout>{10, 7, 0});
+}
```
Drewniok, line 123: Not necessary, but should we maybe add an example where we have two outputs?

Drewniok (dismissed):
Thank you for this new function! It will help us a lot.

include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
```diff
@@ -0,0 +1,192 @@
...
+            return pairwise_distances;
+        };
+
+        const auto dot_distance_comparator = [](const auto& lhs, const auto& rhs) noexcept
+        { return lhs.distance < rhs.distance; };
+
+        // compute pairwise distances
+        auto input_pairwise_distances = compute_pairwise_dot_distances(io_dots);
+        // sort pairwise distances
+        std::sort(input_pairwise_distances.begin(), input_pairwise_distances.end(), dot_distance_comparator);
```
Drewniok, line 132: ``.cbegin``and ``.cend``here?

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
marcelwa: Good catch! That's indeed a leftover from testing.

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
marcelwa: Will do!

Drewniok (dismissed):
LGTM! Thanks for adding more documentation.

marcelwa:
@Drewniok Thinking about this more, it probably makes sense to check for a lower distance threshold as well, since we don't want adjacent dots of, e.g., atomic wires to be paired up, right?

Drewniok:
> @Drewniok Thinking about this more, it probably makes sense to check for a lower distance threshold as well, since we don't want adjacent dots of, e.g., atomic wires to be paired up, right?

@marcelwa That's a good point! I was also worried about this before. However, I thought that this should not be a problem since we have the flags to only collect fro example input cells.

marcelwa:
> @marcelwa That's a good point! I was also worried about this before. However, I thought that this should not be a problem since we have the flags to only collect for example input cells.

Yes, the current algorithm focuses on cells that are flagged with the input label. Two things here, though: (1) Why would we assume that a user would not "misflag" cells or put cells too close together, even if it is accidental? (2) The data type `bdl_pair` is generic and can store arbitrary BDL pairs. Hence, another algorithm can be thought of that collects regular BDL pairs. And that one would also make use of the spacing thresholds.

> I see what you mean. Indeed, for an atomic wire with for example 3 BDL pairs, we would detect 5 pairs, right? (assuming we use 2 nm by default)

I'm not sure if I understand this right. In the current algorithm, each SiDB can only appear in a single pair. It cannot be matched multiple times.

Drewniok:
> > @marcelwa That's a good point! I was also worried about this before. However, I thought that this should not be a problem since we have the flags to only collect for example input cells.
> 
> Yes, the current algorithm focuses on cells that are flagged with the input label. Two things here, though: (1) Why would we assume that a user would not "misflag" cells or put cells too close together, even if it is accidental? (2) The data type `bdl_pair` is generic and can store arbitrary BDL pairs. Hence, another algorithm can be thought of that collects regular BDL pairs. And that one would also make use of the spacing thresholds.
> 

Yes, these are good thoughts. 
1.) That's a good point. Then we should rather go down to 1 nm as default (we should also keep in mind that a look-up table could be useful because the threashold value also depends on the µ value). But I guess it is really hard to provide the correct result for every arbitrary case where the user misflags some DBs.
grafik

Let's take this OR gate where the two input BDLs are a little closer than normal. If a user doesn't mark them correctly, it's really hard to correctly detect the correct input BDL pair.
 
2.) What is the definition of a ``regular BDL pair``?



> > I see what you mean. Indeed, for an atomic wire with for example 3 BDL pairs, we would detect 5 pairs, right? (assuming we use 2 nm by default)
> 
> I'm not sure if I understand this right. In the current algorithm, each SiDB can only appear in a single pair. It cannot be matched multiple times.

Sorry, I was wrong.

marcelwa:
> Yes, these are good thoughts.
> 1.) That's a good point. Then we should rather go down to 1 nm as default (we should also keep in mind that a look-up table could be useful because the threashold value also depends on the µ value). But I guess it is really hard to provide the correct result for every arbitrary case where the user misflags some DBs.

I don't think I understand what you mean by the look-up table idea. Could you elaborate on that, please? Regarding 1 nm as the upper bound: I think that could be too narrow. In the Bestagon gate library, for instance, input/output BDL pairs are ~1.086 nm apart. Given this information, what do you think would be a good upper bound? 1.5 nm maybe?

> 2.) What is the definition of a regular BDL pair?

In my understanding, any two dots that are close enough together such that they are likely to share a charge. In a BDL wire for instance you have always two dots a bit closer together and then a little bit of space and then the next two dots. Those would be regular BDL pairs while the first and last in the row would be input/output BDL pairs, respectively.

Drewniok:
>

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+        set_all_inputs();
+
+        return *this;
+    }
+    /**
+     * Postfix increment operator. Sets the next input state.
+     *
+     * @return Copy of `this` before incrementing.
+     */
+    bdl_input_iterator operator++(int) noexcept
```
Drewniok, line 80: Is there a reason why ``uint64_t`` is not used?

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+        return result;
+    }
+    /**
+     * Addition operator. Computes the input state of the current iterator plus the given integer.
+     *
+     * @param m The amount of input states to skip.
+     * @return The input state of the current iterator plus the given integer.
+     */
+    [[nodiscard]] bdl_input_iterator operator+(const int m) const noexcept
```
Drewniok, line 94: again

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+    }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+    /**
+     * Addition assignment operator. Sets a next input state.
+     *
+     * @param m The amount of input states to skip.
+     * @return Reference to `this`.
+     */
+    bdl_input_iterator& operator+=(const int m) noexcept
```
Drewniok, line 110: again

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+        return *this;
+    }
+#pragma GCC diagnostic pop
+    /**
+     * Subtraction operator. Computes the input state of the current iterator minus the given integer.
+     *
+     * @param m The amount of input states to skip.
+     * @return The input state of the current iterator minus the given integer.
+     */
+    [[nodiscard]] bdl_input_iterator operator-(const int m) const noexcept
```
Drewniok, line 125: again

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+        set_all_inputs();
+
+        return *this;
+    }
+    /**
+     * Postfix decrement operator. Sets the previous input state.
+     *
+     * @return Copy of `this` before decrementing.
+     */
+    bdl_input_iterator operator--(int) noexcept
```
Drewniok, line 151: again

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+    }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-conversion"
+    /**
+     * Subtraction assignment operator. Sets a previous input state.
+     *
+     * @param m The amount of input states to skip.
+     * @return Reference to `this`.
+     */
+    bdl_input_iterator& operator-=(const int m) noexcept
```
Drewniok, line 167: again

include/fiction/algorithms/iter/bdl_input_iterator.hpp
```diff
@@ -0,0 +1,317 @@
...
+        return *this;
+    }
+#pragma GCC diagnostic pop
+    /**
+     * Subscript operator. Computes the input state of the current iterator plus the given integer.
+     *
+     * @param m The amount of input states to skip.
+     * @return The input state of the current iterator plus the given integer.
+     */
+    [[nodiscard]] bdl_input_iterator operator[](const int m) const noexcept
```
Drewniok, line 182: again

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+        REQUIRE(input_bdl_pairs.empty());
+        REQUIRE(output_bdl_pairs.empty());
+        REQUIRE(normal_bdl_pairs.size() == 2);
+
+        const auto& normal_pair1 = normal_bdl_pairs[0];
+        const auto& normal_pair2 = normal_bdl_pairs[1];
+
+        CHECK(normal_pair1.type == layout::cell_type::NORMAL);
+        CHECK(normal_pair1.upper == cell<layout>{2, 0, 0});
```
Drewniok, line 56: I think we cannot assume that normal_pair1 is the "first" normal pair. It could also be ``({3,0,0},{5,0,0})``.

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
```diff
@@ -0,0 +1,301 @@
...
```
Drewniok, line 79: same here

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
```diff
@@ -0,0 +1,301 @@
...
+        REQUIRE(normal_bdl_pairs.size() == 2);
+
+        const auto& input_pair  = input_bdl_pairs.front();
+        const auto& output_pair = output_bdl_pairs.front();
+
+        const auto& normal_pair1 = normal_bdl_pairs[0];
+        const auto& normal_pair2 = normal_bdl_pairs[1];
+
+        CHECK(input_pair.type == layout::cell_type::INPUT);
+        CHECK(input_pair.upper == cell<layout>{0, 0, 0});
```
Drewniok, line 105: same here

test/algorithms/iter/bdl_input_iterator.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    // doing another iteration should overflow and set it back to 0
+    ++bii;
+    CHECK(bii == 2ull);
+
+    const auto& lyt_2 = *bii;
+
+    CHECK(lyt_2.get_cell_type({0, 0, 0}) == sidb_technology::cell_type::INPUT);
+    CHECK(lyt_2.get_cell_type({2, 0, 0}) == sidb_technology::cell_type::EMPTY);
+
```
Drewniok, line 186: What happens if I increment again? Is this something to test?

Drewniok (commented):
Thanks for these terrific additions. It will also make the simulation for different input combinations much easier.

include/fiction/algorithms/iter/bdl_input_iterator.hpp
marcelwa: Because that's how the C++ Standard defines these operators: https://en.cppreference.com/w/cpp/language/operators 

It's the same answer for all of these comments :smile:

test/algorithms/iter/bdl_input_iterator.cpp
marcelwa: We would get to input state 1 again since we have completely wrapped around. This very wrap-around test is already intended to check this particular behavior.

Drewniok: True! Sorry.

test/algorithms/simulation/sidb/detect_bdl_pairs.cpp
marcelwa: I guess if we are truly implementation independent, we cannot assume that indeed. It works because I know how `cell_level_layout` is implemented and that lower indices will be iterated over first. I might change the test.

Drewniok: On my Apple M1 Pro, it actually takes the opposite order. So the test fails.

marcelwa: That's really good to know! Thank you for trying this out. I will adjust the tests accordingly.

test/algorithms/iter/bdl_input_iterator.cpp
marcelwa: No worries :smile:

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,552 @@
...
+     * @return The operational status of the layout under the given simulation parameters.
+     */
+    [[nodiscard]] operational_domain::operational_status
+    is_operational(const sidb_simulation_parameters& sim_params) noexcept
+    {
+        // increment the number of evaluated parameter combinations
+        ++stats.num_evaluated_parameter_combinations;
+
+        // take the first (and only) output BDL pair
+        const auto& output_bdl_pair = output_bdl_pairs.front();
```
Drewniok, line 397: What happens if we have 2-outputs?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,552 @@
...
+            // the expected output of the layout is the i-th bit of the truth table
+            const auto expected_output = kitty::get_bit(truth_table, i);
+
+            // TODO replace with QuickExact
+            // perform an exhaustive ground state simulation
+            const auto sim_result = exhaustive_ground_state_simulation(*bii, sim_params);
+
+            // TODO is this necessary or is it guaranteed that the ground state is always the first element?
+            // find the ground state, which is the charge distribution with the lowest energy
+            const auto ground_state = std::min_element(
```
Drewniok, line 414: You are right! It is required.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,552 @@
...
+                if (charge_state_output_upper != sidb_charge_state::NEGATIVE ||
+                    charge_state_output_lower != sidb_charge_state::NEUTRAL)
+                {
+                    ++stats.num_non_operational_parameter_combinations;
+
+                    return operational_domain::operational_status::NON_OPERATIONAL;
+                }
+            }
+        }
+
```
Drewniok, line 453: At this point, I am a little bit concerned since degeneracy is not covered. I think if there is degeneracy, it should be labeled as ``NON_OPERATIONAL``. Maybe you can take a look at ``energy_distribution.hpp``. It tells you not only the ground state energy, but also its degeneracy.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,552 @@
...
+        // reset the BDL input iterator
+        bii = 0ull;
+
+        // for each input combination
+        for (auto i = 0u; i < truth_table.num_bits(); ++i, ++bii)
+        {
+            // the expected output of the layout is the i-th bit of the truth table
+            const auto expected_output = kitty::get_bit(truth_table, i);
+
```
Drewniok, line 407: As discussed, I think we need a check here like ``is_three_state_simulation_required(*bii)``. After that the question is whether the simulation should be continued with 3-state simulation or whether the layout should be marked as ``NON_OPERATIONAL``. I tend to the latter.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: This is currently ruled out by the assertion: `assert(output_bdl_pairs.size() == 1 && "The layout must have exactly one output BDL pair");`

Mid-term, I want to support multi-output gates, but for now, this suffices.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: Thanks, will do! :+1:

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: Is this function available outside of your QuickExact branch already?

Drewniok: Not yet, but I will generate a PR today.

marcelwa: Many thanks!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: Sorry, I was wrong. Since there can also be degeneracy which does not affect the logic output of the gate, I suggest to just check if all ground states satisfy the logic. Sorry for the wrong information above!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1266 @@
...
+     *
+     * @param samples Maximum number of random samples to be taken before contour tracing.
+     * @return The (partial) operational domain of the layout.
+     */
+    [[nodiscard]] operational_domain contour_tracing(const std::size_t samples) noexcept
+    {
+        mockturtle::stopwatch stop{stats.time_total};
+
+        // first, perform random sampling to find an operational starting point
+        const auto starting_point = find_operational_step_point_via_random_sampling(samples);
```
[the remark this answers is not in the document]
marcelwa: Yes, there is. For now, contour tracing determines a single starting point and works from there. As such, it can abort random sampling as soon as such a starting point is found. Flood fill on the other hand, always performs the number of given random samples in order to potentially find operational points in disconnected domains to expand from. Contour tracing does not yet work with disconnected domains.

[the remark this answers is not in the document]
Drewniok: Makes sense. Thanks.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1265 @@
...
+
+        auto current_contour_point = contour_starting_point;
+        auto backtrack_point       = current_contour_point.x == 0 ?
+                                         current_contour_point :
+                                         step_point{current_contour_point.x - 1, current_contour_point.y};
+
+        auto current_neighborhood = moore_neighborhood(current_contour_point);
+        auto next_point           = next_clockwise_point(current_neighborhood, backtrack_point);
+
+        while (next_point != contour_starting_point)
```
Drewniok, line 469: As you mentioned, contour tracing is not yet applicable to gates where we have disconnected domains. Can it happen that the algorithm never stops? Especially since we can go from one domain to another with ``find_operational_contour_step_point``?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1265 @@
...
+     * @return The Moore neighborhood of the step point at `sp = (x, y)`.
+     */
+    [[nodiscard]] std::vector<step_point> moore_neighborhood(const step_point& sp) const noexcept
+    {
+        std::vector<step_point> neighbors{};
+        neighbors.reserve(8);
+
+        const auto& [x, y] = sp;
+
+        auto decr_x = (x > 0) ? x - 1 : x;
```
Drewniok, line 948: const

include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * @tparam Lyt SiDB cell-level layout type.
+ * @param lyt The layout to detect BDL pairs in.
+ * @param type The type of the SiDBs to detect BDL pairs for, e.g., `INPUT`, `OUTPUT`, `NORMAL`.
+ * @param params Parameters for the BDL pair detection algorithm.
+ * @return A vector of BDL pairs.
+ */
+template <typename Lyt>
+std::vector<bdl_pair<Lyt>> detect_bdl_pairs(const Lyt& lyt, const typename technology<Lyt>::cell_type type,
+                                            const detect_bdl_pairs_params params = {}) noexcept
```
Drewniok, line 95: ``&type``, ``&params``?

include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
```diff
@@ -0,0 +1,267 @@
...
+     */
+    bdl_pair() = default;
+    /**
+     * Constructor for BDL pairs.
+     *
+     * @param t Type of the SiDBs in the pair.
+     * @param u The upper SiDB of the pair.
+     * @param l The lower SiDB of the pair.
+     */
+    bdl_pair(const sidb_technology::cell_type t, const cell<Lyt> u, const cell<Lyt> l) noexcept :
```
Drewniok, line 52: ``&t``, ``&u``, ``&l``?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: Good catch!

include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
Drewniok, line 52: reference for input cells?

include/fiction/algorithms/simulation/sidb/detect_bdl_pairs.hpp
```diff
@@ -0,0 +1,267 @@
...
+             */
+            pairwise_dot_distance() = default;
+            /**
+             * Constructor for pairwise dot distances.
+             *
+             * @param s1 The first dot.
+             * @param s2 The second dot.
+             * @param d The distance between the two dots. (unit: nm).
+             */
+            pairwise_dot_distance(const cell<Lyt> s1, const cell<Lyt> s2, const double d) noexcept :
```
Drewniok, line 140: reference for input cells?

test/algorithms/simulation/sidb/operational_domain.cpp
```diff
@@ -0,0 +1,438 @@
...
+            check_op_domain_params_and_operational_status(op_domain, op_domain_params,
+                                                          operational_domain::operational_status::OPERATIONAL);
+
+            CHECK(mockturtle::to_seconds(op_domain_stats.time_total) > 0.0);
+            CHECK(op_domain_stats.num_simulator_invocations == 200);
+            CHECK(op_domain_stats.num_evaluated_parameter_combinations == 100);
+            CHECK(op_domain_stats.num_operational_parameter_combinations == 100);
+            CHECK(op_domain_stats.num_non_operational_parameter_combinations == 0);
+        }
+        SECTION("contour_tracing")
```
Drewniok, line 144: In rare cases (~10 %), I get ``Assertion failed: (std::find(neighborhood.cbegin(), neighborhood.cend(), backtrack) != neighborhood.cend() && "The backtrack point must be part of the neighborhood"), function operator(), file operational_domain.hpp, line 448.``

test/algorithms/simulation/sidb/operational_domain.cpp
```diff
@@ -0,0 +1,438 @@
...
+        check_op_domain_params_and_operational_status(op_domain, op_domain_params,
+                                                      operational_domain::operational_status::OPERATIONAL);
+
+        CHECK(mockturtle::to_seconds(op_domain_stats.time_total) > 0.0);
+        CHECK(op_domain_stats.num_simulator_invocations == 400);
+        CHECK(op_domain_stats.num_evaluated_parameter_combinations == 100);
+        CHECK(op_domain_stats.num_operational_parameter_combinations == 100);
+        CHECK(op_domain_stats.num_non_operational_parameter_combinations == 0);
+    }
+    SECTION("contour_tracing")
```
Drewniok, line 420: also here the same as above.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1265 @@
...
+            }
+
+            // if no physically-valid charge distributions were found, the layout is non-operational
+            if (sim_result.charge_distributions.empty())
+            {
+                return non_operational();
+            }
+
+            // find the ground state, which is the charge distribution with the lowest energy
+            const auto ground_state = std::min_element(
```
Drewniok, line 806: If we have a layout with a degenerated ground state, this could cause some problems in rare cases (the computation could be even hardware dependent).

test/algorithms/simulation/sidb/operational_domain.cpp
marcelwa: Huh curious :thinking: Will investigate. Thank you!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: What would be your suggestion?

Drewniok: I think there are a few ways to do this. Maybe one way to do it:
```suggestion
            // ground state is degenerate
            if (energy_distribution(sim_result.charge_distributions)[0]>1)
            {
                return non_operational();
            }
```

marcelwa: Ok, I could do that. However, I still need the ground state to work with in the following lines.

test/algorithms/simulation/sidb/operational_domain.cpp
marcelwa: I've found the source of the problem and fixed it. Many thanks for reporting! :pray:

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1267 @@
...
+    }
+};
+
+}  // namespace detail
+
+/**
+ * Computes the operational domain of the given SiDB cell-level layout. The operational domain is the set of all
+ * parameter combinations for which the layout is logically operational. Logical operation is defined as the layout
+ * implementing the given truth table. The input BDL pairs of the layout are assumed to be in the same order as the
+ * inputs of the truth table. // TODO implement the matching of truth table inputs and BDL pair ordering
```
Drewniok, line 1056: Just so we don't forget it at the end: We have to delete the ``TODO``.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1267 @@
...
+        *stats = st;
+    }
+
+    return result;
+}
+/**
+ * Computes the operational domain of the given SiDB cell-level layout. The operational domain is the set of all
+ * parameter combinations for which the layout is logically operational. Logical operation is defined as the layout
+ * implementing the given truth table. The input BDL pairs of the layout are assumed to be in the same order as the
+ * inputs of the truth table. // TODO implement the matching of truth table inputs and BDL pair ordering
```
Drewniok, line 1142: here is also a ``TODO``

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1260 @@
...
+
+  private:
+    /**
+     * The SiDB cell-level layout to investigate.
+     */
+    const Lyt& layout;
+    /**
+     * The specification of the layout.
+     */
+    const TT& truth_table;  // TODO implement the matching of multi-input truth table inputs and BDL pair ordering
```
Drewniok, line 505: here is a ``TODO``

marcelwa: Yes, this is indeed still a TODO for the extension. For now, it can be there as it doesn't show up in the documentation. And I don't want to forget it ;) And the code works without it as is.

Drewniok: Perfect!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -0,0 +1,1260 @@
...
+
+            // if no physically-valid charge distributions were found, the layout is non-operational
+            if (sim_result.charge_distributions.empty())
+            {
+                return non_operational();
+            }
+
+            // if the ground state is degenerate, the layout is non-operational
+            if (const auto energy_distr = energy_distribution(sim_result.charge_distributions);
+                energy_distr.count(0.0) > 0)
```
Drewniok, line 818: @marcelwa This checks if the key ``0.0`` exists, right? If so, this would be wrong, because the ground state energy is different and only zero if all SiDBs are neutrally charged.

include/fiction/io/csv_writer.hpp
```diff
@@ -5,58 +5,86 @@
...
 
 namespace fiction
 {
 
+/**
+ * Utility class for writing delimited (e.g. CSV) data into an output stream. It provides a variadic member function,
+ * `write_line`, that can take an arbitrary number of arguments and write them to the provided output stream in a line
+ * separated by a specified delimiter.
+ *
+ * The csv_writer follows some behavior principles:
```
Drewniok, line 19: @marcelwa Just a small comment: Do you really want those details?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: Thanks, should be fixed now :+1:

Drewniok (dismissed):
Many thanks, LGTM! This is a super cool and convienent contribution.

experiments/operational_domain/operational_domain_siqad.cpp
```diff
@@ -0,0 +1,187 @@
...
+            // update the total number of simulator calls
+            total_sim_calls_gs += op_domain_stats_gs.num_simulator_invocations;
+            total_sim_calls_rs += op_domain_stats_rs.num_simulator_invocations;
+            total_sim_calls_ff += op_domain_stats_ff.num_simulator_invocations;
+            total_sim_calls_ct += op_domain_stats_ct.num_simulator_invocations;
+
+            // compute the operational percentages
+            const auto operational_percentage_gs =
+                static_cast<double>(op_domain_stats_gs.num_operational_parameter_combinations) /
+                op_domain_stats_gs.num_evaluated_parameter_combinations;
```
Drewniok, line 130: My clang-tidy shows an issue because we convert ``std::size_t`` to ``double``

experiments/operational_domain/operational_domain_siqad.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include "fiction/io/write_operational_domain.hpp"                    // writer for operational domains
+#include "fiction/types.hpp"                                          // pre-defined types suitable for the FCN domain
+#include "fiction/utils/truth_table_utils.hpp"                        // truth tables helper functions
+#include "fiction_experiments.hpp"                                    // experiment class
+
+#include <fmt/format.h>                   // string formatting
+#include <kitty/dynamic_truth_table.hpp>  // truth tables
+
+#include <array>
+#include <cstdint>
```
Drewniok, line 16: We don't need this, right?

experiments/operational_domain/operational_domain_siqad.cpp
```diff
@@ -0,0 +1,187 @@
...
+            const auto operational_percentage_ff =
+                static_cast<double>(op_domain_stats_ff.num_operational_parameter_combinations) /
+                op_domain_stats_ff.num_evaluated_parameter_combinations;
+            const auto operational_percentage_ct =
+                static_cast<double>(op_domain_stats_ct.num_operational_parameter_combinations) /
+                op_domain_stats_ct.num_evaluated_parameter_combinations;
+
+            opdomain_exp(
+                // Benchmark
+                benchmark.string(), lyt.num_cells(),
```
Drewniok, line 143: The only disadvantage of ``benchmark.string()`` is that we get the whole path, right? This can be quite long.

experiments/operational_domain/operational_domain_siqad.cpp
marcelwa: Yes, we do. We use `std::size_t` and `uint64_t`.

experiments/operational_domain/operational_domain_siqad.cpp
marcelwa: I don't mind it as it prevents confusion if you have similar files in different directories.

experiments/operational_domain/operational_domain_siqad.cpp
marcelwa: Good catch!

experiments/operational_domain/operational_domain_siqad.cpp
Drewniok: @marcelwa Absolutely! It is less error-prone.

marcelwa:
`clang-tidy` does not make it through this PR anymore in 6 hours, which is the maximum runtime we get on free GitHub runners. Therefore, I will merge this now regardless of the failing CI. There is not much we can do about it, unfortunately. As we saw before, it is mostly clean anyway.

PR #325, opened 2023-11-05
:zap: Collection of optimizations and a bug fix for `charge_distribution_surface`

wlambooy:
## Description

This PR introduces a collection of optimisations, mainly with regard to `charge_distribution_surface`.
The commit messages describe the changes that are made.

Most prominently, we have:
- Fixed a bug where if the charge index is changed from i to i', and i' has more leading zeroes than i, then there would be cells that stay unchanged that need to be assigned a negative charge state.
- The system energy is now by default only recomputed for physically valid charge distributions.
- The `validity_check` function now returns as soon as a verdict can be made.
- Functions that modify the private storage object can be made `const` apparently. This may be applicable in other files as well. Currently I applied the change in `charge_distribution_surface.hpp` and `cell_level_layout.hpp`.
- Removed an unnecessary mutex lock in `quicksim.hpp`. Also the `update_after_charge_change()` call before calling the threads can be omitted, since in each thread we first make such a call before checking physical validity. Hence also we have a proper initialisation of the local potentials and system energy before calling `adjacent_search`.

## Checklist:

### DIFF
--- a/include/fiction/algorithms/simulation/sidb/quickexact.hpp
+++ b/include/fiction/algorithms/simulation/sidb/quickexact.hpp
@@ -268,6 +268,14 @@ class quickexact_impl
         static_assert(has_siqad_coord_v<ChargeLyt>, "ChargeLyt is not based on SiQAD coordinates");
         static_assert(is_charge_distribution_surface_v<ChargeLyt>, "ChargeLyt is not a charge distribution surface");
 
+        if (base_number == required_simulation_base_number::THREE)
+        {
+            charge_lyt.assign_base_number(3);
+        }
+        else
+        {
+            charge_lyt.assign_base_number(2);
+        }
         charge_layout.assign_physical_parameters(params.physical_parameters);
         charge_layout.assign_all_charge_states(sidb_charge_state::NEUTRAL);
         charge_layout.assign_dependent_cell(all_sidbs_in_lyt_without_negative_preassigned_ones[0]);
--- a/include/fiction/algorithms/simulation/sidb/quickexact.hpp
+++ b/include/fiction/algorithms/simulation/sidb/quickexact.hpp
@@ -99,6 +99,11 @@ class quickexact_impl
         result.algorithm_name      = "QuickExact";
         result.physical_parameters = params.physical_parameters;
 
+        if (layout.num_cells() == 0)
+        {
+            return result;
+        }
+
         mockturtle::stopwatch<>::duration time_counter{};
         {
             const mockturtle::stopwatch stop{time_counter};
--- a/include/fiction/algorithms/simulation/sidb/quickexact.hpp
+++ b/include/fiction/algorithms/simulation/sidb/quickexact.hpp
@@ -86,7 +86,7 @@ class quickexact_impl
 {
   public:
     quickexact_impl(const Lyt& lyt, const quickexact_params<Lyt>& parameter) :
-            layout{lyt},
+            layout{lyt.clone()},
             charge_lyt{lyt},
             params{parameter}
     {
@@ -349,22 +349,14 @@ class quickexact_impl
 
             if (charge_layout.is_physically_valid())
             {
-                charge_distribution_surface<Lyt> charge_lyt_copy{charge_layout};
+                charge_distribution_surface<Lyt> charge_lyt_copy{charge_lyt};
+
+                charge_layout.foreach_cell(
+                    [&charge_lyt_copy, &charge_layout](const auto& c)
+                    { charge_lyt_copy.assign_charge_state(c, charge_layout.get_charge_state(c)); });
+
+                charge_lyt_copy.update_after_charge_change();
                 charge_lyt_copy.recompute_system_energy();
-
-                // The pre-assigned negatively-charged SiDBs are added to the final layout.
-                for (const auto& cell : preassigned_negative_sidbs)
-                {
-                    charge_lyt_copy.add_sidb(cell, sidb_charge_state::NEGATIVE);
-                }
-
-                if constexpr (has_get_sidb_defect_v<Lyt>)
-                {
-                    for (const auto& [cell, defect] : real_placed_defects)
-                    {
-                        charge_lyt_copy.assign_sidb_defect(cell, defect);
-                    }
-                }
                 result.charge_distributions.push_back(charge_lyt_copy);
             }
         }
@@ -402,23 +394,14 @@ class quickexact_impl
             {
                 if (charge_layout.is_physically_valid())
                 {
-                    charge_distribution_surface<Lyt> charge_lyt_copy{charge_layout};
+                    charge_distribution_surface<Lyt> charge_lyt_copy{charge_lyt};
+
+                    charge_layout.foreach_cell(
+                        [&charge_lyt_copy, &charge_layout](const auto& c)
+                        { charge_lyt_copy.assign_charge_state(c, charge_layout.get_charge_state(c)); });
+
+                    charge_lyt_copy.update_after_charge_change();
                     charge_lyt_copy.recompute_system_energy();
-
-                    // The pre-assigned negatively-charged SiDBs are added to the final layout.
-                    for (const auto& cell : preassigned_negative_sidbs)
-                    {
-                        charge_lyt_copy.add_sidb(cell, sidb_charge_state::NEGATIVE);
-                    }
-
-                    if constexpr (has_get_sidb_defect_v<Lyt>)
-                    {
-                        for (const auto& [cell, defect] : real_placed_defects)
-                        {
-                            charge_lyt_copy.assign_sidb_defect(cell, defect);
-                        }
-                    }
-
                     result.charge_distributions.push_back(charge_lyt_copy);
                 }
 
@@ -432,22 +415,14 @@ class quickexact_impl
 
             if (charge_layout.is_physically_valid())
             {
-                charge_distribution_surface<Lyt> charge_lyt_copy{charge_layout};
+                charge_distribution_surface<Lyt> charge_lyt_copy{charge_lyt};
+
+                charge_layout.foreach_cell(
+                    [&charge_lyt_copy, &charge_layout](const auto& c)
+                    { charge_lyt_copy.assign_charge_state(c, charge_layout.get_charge_state(c)); });
+
+                charge_lyt_copy.update_after_charge_change();
                 charge_lyt_copy.recompute_system_energy();
-
-                for (const auto& cell : preassigned_negative_sidbs)
-                {
-                    charge_lyt_copy.add_sidb(cell, sidb_charge_state::NEGATIVE);
-                }
-
-                if constexpr (has_get_sidb_defect_v<Lyt>)
-                {
-                    for (const auto& [cell, defect] : real_placed_defects)
-                    {
-                        charge_lyt_copy.assign_sidb_defect(cell, defect);
-                    }
-                }
-
                 result.charge_distributions.push_back(charge_lyt_copy);
             }
 
@@ -469,23 +444,15 @@ class quickexact_impl
         {
             if (charge_layout.is_physically_valid())
             {
-                charge_distribution_surface<Lyt> charge_lyt_copy{charge_layout};
+                charge_distribution_surface<Lyt> charge_lyt_copy{charge_lyt};
+
+                charge_layout.foreach_cell(
+                    [&charge_lyt_copy, &charge_layout](const auto& c)
+                    { charge_lyt_copy.assign_charge_state(c, charge_layout.get_charge_state(c)); });
+
+                charge_lyt_copy.update_after_charge_change();
                 charge_lyt_copy.recompute_system_energy();
-
-                // The pre-assigned negatively-charged SiDBs are added to the final layout.
-                for (const auto& cell : preassigned_negative_sidbs)
-                {
-                    charge_lyt_copy.add_sidb(cell, sidb_charge_state::NEGATIVE);
-                }
-
-                if constexpr (has_get_sidb_defect_v<Lyt>)
-                {
-                    for (const auto& [cell, defect] : real_placed_defects)
-                    {
-                        charge_lyt_copy.assign_sidb_defect(cell, defect);
-                    }
-                }
-
+                charge_lyt_copy.charge_distribution_to_index_general();
                 result.charge_distributions.push_back(charge_lyt_copy);
             }
 
@@ -496,21 +463,14 @@ class quickexact_impl
 
         if (charge_layout.is_physically_valid())
         {
-            charge_distribution_surface<Lyt> charge_lyt_copy{charge_layout};
+            charge_distribution_surface<Lyt> charge_lyt_copy{charge_lyt};
 
-            for (const auto& cell : preassigned_negative_sidbs)
-            {
-                charge_lyt_copy.add_sidb(cell, sidb_charge_state::NEGATIVE);
-            }
+            charge_layout.foreach_cell([&charge_lyt_copy, &charge_layout](const auto& c)
+                                       { charge_lyt_copy.assign_charge_state(c, charge_layout.get_charge_state(c)); });
 
-            if constexpr (has_get_sidb_defect_v<Lyt>)
-            {
-                for (const auto& [cell, defect] : real_placed_defects)
-                {
-                    charge_lyt_copy.assign_sidb_defect(cell, defect);
-                }
-            }
-
+            charge_lyt_copy.update_after_charge_change();
+            charge_lyt_copy.recompute_system_energy();
+            charge_lyt_copy.charge_distribution_to_index_general();
             result.charge_distributions.push_back(charge_lyt_copy);
         }
 
--- a/include/fiction/algorithms/simulation/sidb/quickexact.hpp
+++ b/include/fiction/algorithms/simulation/sidb/quickexact.hpp
@@ -199,10 +199,6 @@ class quickexact_impl
                     result.charge_distributions.push_back(charge_lyt_copy);
                 }
             }
-            else
-            {
-                result.charge_distributions.push_back(charge_lyt);
-            }
 
             for (const auto& cell : preassigned_negative_sidbs)
             {
--- a/include/fiction/algorithms/simulation/sidb/quickexact.hpp
+++ b/include/fiction/algorithms/simulation/sidb/quickexact.hpp
@@ -199,6 +199,10 @@ class quickexact_impl
                     result.charge_distributions.push_back(charge_lyt_copy);
                 }
             }
+            else
+            {
+                result.charge_distributions.push_back(charge_lyt);
+            }
 
             for (const auto& cell : preassigned_negative_sidbs)
             {
--- a/include/fiction/algorithms/simulation/sidb/quicksim.hpp
+++ b/include/fiction/algorithms/simulation/sidb/quicksim.hpp
@@ -93,7 +93,7 @@ sidb_simulation_result<Lyt> quicksim(const Lyt& lyt, const quicksim_params& ps =
         charge_lyt.assign_base_number(2);
         charge_lyt.assign_all_charge_states(sidb_charge_state::NEGATIVE);
         charge_lyt.update_after_charge_change(dependent_cell_mode::VARIABLE);
-        const auto negative_sidb_indices = charge_lyt.negative_sidb_detection();
+        const auto& negative_sidb_indices = charge_lyt.negative_sidb_detection();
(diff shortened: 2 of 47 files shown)


wlambooy:
Looks like I missed some tests still. Probably just some silliness

Drewniok:
@wlambooy Thank you for your efforts! As soon as it passes the tests, I will take a closer look at it. Please note that ``const`` indicates to the compiler that the function is a read-only operation.

wlambooy:
> @wlambooy Thank you for your efforts! As soon as it passes the tests, I will take a closer look at it. Please note that `const` indicates to the compiler that the function is a read-only operation.

I missed some consequences of making the empty layout physically valid, hopefully everything passes now.
My intuition is to use keywords like `const` as much as possible to help the compiler, since the typing system makes sure you can't overdo it. Yet I would think that a function that modifies a private object cannot be const, yet it is accepted. Do you understand why?

Drewniok:
I see what you mean, but this is not the best style. Using the keyword ``const`` when the member function obviously changes the logical state of the object violates semantic clarity and logical constness. I would ask you to only use/add const where it is really appropriate. Thank you!

wlambooy:
> I see what you mean, but this is not the best style. Using the keyword `const` when the member function obviously changes the logical state of the object violates semantic clarity and logical constness. I would ask you to only use/add const where it is really appropriate. Thank you!

My assumption that the C++ compiler catches this turns out to be incorrect, and our example in which a dereferenced pointer to a object is changed show this. I made sure to verify the logical constness in a new commit.

While trying to improve the code coverage, I stumbled on a layout that generates no charge distributions for ExGS, but two for QuickExact. The test looks as follows:
```c++
TEMPLATE_TEST_CASE(
    "ExGS simulation of two SiDBs placed directly next to each other with non-realistic relative permittivity", "[exhaustive-ground-state-simulation]",
    (cell_level_layout<sidb_technology, clocked_layout<cartesian_layout<siqad::coord_t>>>),
    (charge_distribution_surface<cell_level_layout<sidb_technology, clocked_layout<cartesian_layout<siqad::coord_t>>>>))
{
    TestType lyt{};
    lyt.assign_cell_type({1, 3, 0}, TestType::cell_type::NORMAL);
    lyt.assign_cell_type({2, 3, 0}, TestType::cell_type::NORMAL);

    const quickexact_params<TestType> params{sidb_simulation_parameters{2, -0.32, 1.0e-3}};

    const auto simulation_results = quickexact<TestType>(lyt, params);

    CHECK(simulation_results.charge_distributions.empty());
}
```

A degenerate charge distribution is returned by QuickExact with a system energy of -3472.6959461051524. Perhaps QuickExact is not defined on such non-realistic relative permittivity values?

Drewniok:
> > I see what you mean, but this is not the best style. Using the keyword `const` when the member function obviously changes the logical state of the object violates semantic clarity and logical constness. I would ask you to only use/add const where it is really appropriate. Thank you!
> 
> My assumption that the C++ compiler catches this turns out to be incorrect, and our example in which a dereferenced pointer to a object is changed show this. I made sure to verify the logical constness in a new commit.
> 
> While trying to improve the code coverage, I stumbled on a layout that generates no charge distributions for ExGS, but two for QuickExact. The test looks as follows:
> 
> ```c++
> TEMPLATE_TEST_CASE(
>     "ExGS simulation of two SiDBs placed directly next to each other with non-realistic relative permittivity", "[exhaustive-ground-state-simulation]",
>     (cell_level_layout<sidb_technology, clocked_layout<cartesian_layout<siqad::coord_t>>>),
>     (charge_distribution_surface<cell_level_layout<sidb_technology, clocked_layout<cartesian_layout<siqad::coord_t>>>>))
> {
>     TestType lyt{};
>     lyt.assign_cell_type({1, 3, 0}, TestType::cell_type::NORMAL);
>     lyt.assign_cell_type({2, 3, 0}, TestType::cell_type::NORMAL);
> 
>     const quickexact_params<TestType> params{sidb_simulation_parameters{2, -0.32, 1.0e-3}};
> 
>     const auto simulation_results = quickexact<TestType>(lyt, params);
> 
>     CHECK(simulation_results.charge_distributions.empty());
> }
> ```
> 
> A degenerate charge distribution is returned by QuickExact with a system energy of -3472.6959461051524. Perhaps QuickExact is not defined on such non-realistic relative permittivity values?

Thank you for your efforts! 🙏 
That's a good example, even though it's not realistic. You are right. I found the problem and will fix it as soon as possible. 
Just for your information: Be aware of this additional parameter, which can also lead to deviations between both simulators, but not here.
```cpp
    const quickexact_params<TestType> params{sidb_simulation_parameters{2, -0.32, 1.0e-3},
                                             quickexact_params<TestType>::automatic_base_number_detection::OFF};
```

include/fiction/algorithms/simulation/sidb/quickexact.hpp
```diff
@@ -199,6 +199,10 @@ class quickexact_impl
                     result.charge_distributions.push_back(charge_lyt_copy);
                 }
             }
+            else
```
Drewniok, line 202: Is this important to you in any way? I am not a big fan of returning an empty charge distribution.

include/fiction/technology/charge_distribution_surface.hpp
```diff
@@ -609,7 +614,7 @@ class charge_distribution_surface<Lyt, false> : public Lyt
      *
      * @return Vector of indices describing which SiDBs must be negatively charged.
      */
-    std::vector<int64_t> negative_sidb_detection() noexcept
+    std::vector<int64_t> negative_sidb_detection() const noexcept
```
Drewniok, line 617: without ``const``

test/algorithms/simulation/sidb/exhaustive_ground_state_simulation.cpp
```diff
@@ -25,7 +25,7 @@ TEMPLATE_TEST_CASE(
 
     const auto simulation_results = exhaustive_ground_state_simulation<TestType>(lyt, params);
 
-    CHECK(simulation_results.charge_distributions.empty());
+    CHECK(simulation_results.charge_distributions.size() == 1);
```
Drewniok, line 28: I am not a big fan that we have an entry here as mentioned above.

test/algorithms/simulation/sidb/quickexact.cpp
```diff
@@ -27,7 +27,7 @@ TEMPLATE_TEST_CASE(
 
     const auto simulation_results = quickexact(lyt, params);
 
-    CHECK(simulation_results.charge_distributions.empty());
+    CHECK(simulation_results.charge_distributions.size() == 1);
```
Drewniok, line 30: same

test/algorithms/simulation/sidb/quicksim.cpp
```diff
@@ -43,14 +43,13 @@ TEMPLATE_TEST_CASE(
 
     const auto simulation_results = quicksim<TestType>(lyt, quicksim_params);
 
-    CHECK(simulation_results.charge_distributions.empty());
+    CHECK(simulation_results.charge_distributions.size() == 2);
```
Drewniok, line 46: Especially in this case, it is super counterintuitive to have two entries even though an empty layout was simulated. I would highly recommend keeping it the way it was.

include/fiction/algorithms/simulation/sidb/quickexact.hpp
wlambooy: I noticed `is_physically_valid` performs a check for the empty layout in each call which seems illogical as this function is called so often during simulation which performed on non-empty layouts in practice. Would you agree to that? I could also just make a case distinction for the empty layout in the simulators (and return no valid charge distributions) instead of this remedy

include/fiction/technology/charge_distribution_surface.hpp
wlambooy: I purposefully left this when removing the logically non-const `const` type specifiers. It looks to me that it does not change the state of the `charge_distribution_surface` object or call a function that does. If you intended the function to imply some kind of non-constness I'd be interested to know :)

Drewniok:
@wlambooy Thank you very much for your effort! Just klick on ``re-request review`` when you are ready.

Drewniok (approved):
LGTM! Thank you so much for your efforts! 🙏

marcelwa:
@wlambooy many thanks also from my side for the additions! :pray:

@Drewniok thanks to you as well for taking your time reviewing the changes! I'll have a quick scan as well and then it should be ready for merge :muscle:

include/fiction/algorithms/simulation/sidb/quicksim.hpp
```diff
@@ -78,6 +78,12 @@ sidb_simulation_result<Lyt> quicksim(const Lyt& lyt, const quicksim_params& ps =
     st.additional_simulation_parameters.emplace_back("iteration_steps", ps.interation_steps);
     st.additional_simulation_parameters.emplace_back("alpha", ps.alpha);
     st.physical_parameters = ps.phys_params;
+
+    if (lyt.num_cells() == 0)
+    {
+        return st;
+    }
+
     st.charge_distributions.reserve(ps.interation_steps);
```
marcelwa, line 87: I know you didn't name this variable `st` but now that there have been changes made, it seems to be an inconsistency to the other two ground state algorithms where this variable is called `result`, a more fitting name in my opinion. May I ask you to adjust the naming? Many thanks!

include/fiction/algorithms/simulation/sidb/quicksim.hpp
```diff
@@ -93,7 +99,7 @@ sidb_simulation_result<Lyt> quicksim(const Lyt& lyt, const quicksim_params& ps =
         charge_lyt.assign_base_number(2);
         charge_lyt.assign_all_charge_states(sidb_charge_state::NEGATIVE);
         charge_lyt.update_after_charge_change(dependent_cell_mode::VARIABLE);
-        const auto negative_sidb_indices = charge_lyt.negative_sidb_detection();
+        const auto& negative_sidb_indices = charge_lyt.negative_sidb_detection();
```
marcelwa, line 102: Why the reference? Due to C++'s return value optimization, it should be equivalent performance-wise to have `const auto` instead of `const auto&`. However, `const auto` is safer because it doesn't fall prey to the object running out of scope and disappearing. I prefer using `const auto` unless there are good reasons against it.

include/fiction/algorithms/simulation/sidb/time_to_solution.hpp
```diff
@@ -105,14 +105,14 @@ void time_to_solution(Lyt& lyt, const quicksim_params& quicksim_params, const ti
     sidb_simulation_result<Lyt> simulation_result{};
     if (tts_params.engine == exhaustive_sidb_simulation_engine::QUICKEXACT)
     {
+        const quickexact_params<Lyt> params{quicksim_params.phys_params};
         st.algorithm      = "QuickExact";
```
marcelwa, line 109: Should `st` also be `result` here?

include/fiction/algorithms/simulation/sidb/time_to_solution.hpp
```diff
@@ -105,14 +105,14 @@ void time_to_solution(Lyt& lyt, const quicksim_params& quicksim_params, const ti
...
-        simulation_result = exhaustive_ground_state_simulation(lyt, quicksim_params.phys_params);
+        simulation_result = quickexact(lyt, params);
     }
     else
     {
-        const quickexact_params<Lyt> params{quicksim_params.phys_params};
-        st.algorithm      = "QuickExact";
-        simulation_result = quickexact(lyt, params);
+        st.algorithm      = "Exhaustive Ground State Simulation";
```
marcelwa, line 114: I believe we use `ExGS` as the algorithm name here. @Drewniok can you confirm this?

include/fiction/technology/charge_distribution_surface.hpp
```diff
@@ -329,28 +334,28 @@ class charge_distribution_surface<Lyt, false> : public Lyt
...
-            this->update_local_potential();
-            this->recompute_system_energy();
-            this->validity_check();
         }
-        else
-        {
-            strg->phys_params = params;
+
+        if (old_params.lat_a != params.lat_a || old_params.lat_b != params.lat_b || old_params.lat_c != params.lat_c)
+        {  // lattice changed
```
marcelwa, line 347: Either move the comment directly above the `if` statement (if it concerns the condition) or within the block (if it concerns the statements inside the condition).

include/fiction/technology/charge_distribution_surface.hpp
```diff
@@ -1966,59 +1962,58 @@ class charge_distribution_surface<Lyt, false> : public Lyt
     }
     /**
      *  The stored unique index is converted to a charge distribution.
+     *
+     *  @param incremented Flag that can be set to true when the charge index was incremented before calling this
+     * function to enable an optimization. When set to true, the leading zeroes of the charge index are not handled.
      */
-    void index_to_charge_distribution() noexcept
+    void index_to_charge_distribution(const bool incremented = false) noexcept
```
marcelwa, line 1969: Instead of `bool` arguments, we usually prefer a dedicated `enum class` that provides more meaning to the arguments. From the users' perspective, there is not much meaning in a statement like `index_to_charge_distribution(false);`. However, `index_to_charge_distribution(charge_index_recomputation::FROM_SCRATCH);` (or similar) is way more intuitive.

include/fiction/technology/charge_distribution_surface.hpp
```diff
@@ -1966,59 +1962,58 @@ class charge_distribution_surface<Lyt, false> : public Lyt
     }
     /**
      *  The stored unique index is converted to a charge distribution.
+     *
+     *  @param incremented Flag that can be set to true when the charge index was incremented before calling this
+     * function to enable an optimization. When set to true, the leading zeroes of the charge index are not handled.
```
marcelwa, line 1967:
```suggestion
     *  @param incremented Flag that can be set to `true` if the charge index was incremented right before calling this
     * function to enable an optimization. When set to `true`, the leading zeroes of the charge index are ignored.
```

marcelwa: Please adjust accordingly if you agree to change the argument type of this function.

marcelwa (changes requested):

include/fiction/algorithms/simulation/sidb/time_to_solution.hpp
Drewniok: @marcelwa You are right!
@wlambooy Please use ``ExGS``. Thanks

marcelwa:
> @wlambooy Thanks again for your efforts! @marcelwa just asked me to check again if your changes do not increase the simulation runtimes. However, when running our recently added benchmark script (fiction -> test -> benchmark -> simulation.cpp), a careful analysis of the results shows that the runtimes increase significantly (~10%), which is unacceptable. As far as I can see, this is not caused by the bugfix, but rather by the other changes.

"Unacceptable" is maybe a bit harsh a statement. We aim for the most performant code possible for any given scenario and we indeed take pride in our achievements. @wlambooy if the benchmark code @Drewniok mentioned helps you figure out where exactly the runtime increase was caused, we'd appreciate a fix. Many thanks!

marcelwa:
> > Also, @Drewniok, can I ask what your benchmark analysis process looks like so that I can replicate it? I see the simulation benchmark file contains a comment with `mean`, `low mean` and `high mean` columns, are these the results from 3 separate benchmark runs, or are they the mean/low/high from a set of more benchmark runs?
> 
> @wlambooy See https://github.com/catchorg/Catch2/blob/devel/docs/benchmarks.md for a definition of samples, iterations, etc. In our case, the crossing gate will be simulated 100 times due to 100 samples and only 1 iteration each.

I think @wlambooy's question was where the output comes from. It will be automatically generated by the benchmark script. Simply pass `-DFICTION_BENCHMARK=ON` to your `CMake` call and run the resulting binary after compilation in `Release` mode.

wlambooy:
> > Also, @Drewniok, can I ask what your benchmark analysis process looks like so that I can replicate it? I see the simulation benchmark file contains a comment with `mean`, `low mean` and `high mean` columns, are these the results from 3 separate benchmark runs, or are they the mean/low/high from a set of more benchmark runs?
> 
> @wlambooy See https://github.com/catchorg/Catch2/blob/devel/docs/benchmarks.md for a definition of samples, iterations, etc. In our case, the crossing gate will be simulated 100 times due to 100 samples and only 1 iteration each.

I see, I was running the benchmarks from within CLion so I was getting some XML output that didn't look like the output in the comment. I have to say that on my system at least, I'm always getting quite a spread in the benchmark results, and usually the more often I run it without rebuilding the better the times will get. You can see this in the benchmark results below, which were performed directly in sequence (look at the `mean value`s)

```
QuickExact
<!-- All values in nano seconds -->
<mean value="2.42783e+09" lowerBound="2.35123e+09" upperBound="2.54392e+09" ci="0.95"/>
<standardDeviation value="4.74379e+08" lowerBound="3.47914e+08" upperBound="6.37069e+08" ci="0.95"/>
<outliers variance="0.936384" lowMild="0" lowSevere="0" highMild="2" highSevere="7"/>

QuickSim
<!-- All values in nano seconds -->
<mean value="1.33811e+07" lowerBound="1.31979e+07" upperBound="1.35853e+07" ci="0.95"/>
<standardDeviation value="987767" lowerBound="835378" upperBound="1.19855e+06" ci="0.95"/>
<outliers variance="0.676348" lowMild="2" lowSevere="0" highMild="6" highSevere="1"/>


QuickExact
<!-- All values in nano seconds -->
<mean value="2.11722e+09" lowerBound="2.08929e+09" upperBound="2.15057e+09" ci="0.95"/>
<standardDeviation value="1.55694e+08" lowerBound="1.30843e+08" upperBound="1.96559e+08" ci="0.95"/>
<outliers variance="0.666639" lowMild="0" lowSevere="0" highMild="3" highSevere="0"/>

QuickSim
<!-- All values in nano seconds -->
<mean value="1.06471e+07" lowerBound="1.04801e+07" upperBound="1.08103e+07" ci="0.95"/>
<standardDeviation value="845190" lowerBound="739383" upperBound="995966" ci="0.95"/>
<outliers variance="0.707146" lowMild="2" lowSevere="0" highMild="1" highSevere="0"/>


QuickExact
<!-- All values in nano seconds -->
<mean value="2.01781e+09" lowerBound="2.00066e+09" upperBound="2.04291e+09" ci="0.95"/>
<standardDeviation value="1.04298e+08" lowerBound="7.61848e+07" upperBound="1.52602e+08" ci="0.95"/>
<outliers variance="0.494789" lowMild="0" lowSevere="0" highMild="3" highSevere="2"/>

Quicksim
<!-- All values in nano seconds -->
<mean value="9.83447e+06" lowerBound="9.68097e+06" upperBound="9.98761e+06" ci="0.95"/>
<standardDeviation value="783977" lowerBound="703335" upperBound="901620" ci="0.95"/>
<outliers variance="0.707254" lowMild="0" lowSevere="0" highMild="0" highSevere="0"/>
```

wlambooy:
After doing a couple of benchmark runs with pre-built binaries in a more processor-usage sterile environement, I am able to confirm a performance difference with the following findings:

The QuickExact time is _slower_ (+5.4% runtime on average) while QuickSim is _faster_ (-7.1% runtime on average). All benchmark runs were consistent in the differences: each QuickExact benchmark run of the main branch was faster than each QuickExact benchmark run of this branch, and each QuickSim benchmark run of the main branch was slower than each QuickSim benchmark run of this branch.

I'll try to figure out why this is.

Drewniok:
@wlambooy Thanks again for your efforts! I know, sometimes it is not so easy to find out why the code is slower. Take your time and don't hesitate to contact us. 
Since it would be great to have this bugfix in the ``main`` branch as soon as possible, can I ask you to maybe split this PR into two - one for the bugfix and one for the other changes. What do you think about that? What do you think @marcelwa?

include/fiction/technology/charge_distribution_surface.hpp
wlambooy: #347 implements this change

include/fiction/technology/charge_distribution_surface.hpp
wlambooy: #347 implements this change

wlambooy:
I was working on more pressing things the last week so I haven't gotten into a code speed investigation. I put the bugfixes in a new PR: #347

marcelwa:
This ran out of sync and doesn't seem to be necessary anymore according to @Drewniok.

PR #326, opened 2023-11-07
🐛 The dependent cell is now linked correctly to the simulation base number in SiDB simulation

Drewniok:
## Description

This PR fixes an issue where the dependent cell can be positively charged even if the base number is set to two-state simulation.

@wlambooy Thanks for reporting this bug.

## Checklist:


marcelwa (approved):

marcelwa:
@Drewniok thanks for fixing this so quickly!

Would you mind changing the name of the PR slightly? It should not state what was previously the case but what you have implemented, e.g., "🐛 The dependent cell is now linked correctly to the simulation base number in QuickExact".

marcelwa:
@Drewniok I have no idea why the `clang-tidy` action seems to be running forever. I don't think it's your fault.

PR #349, opened 2023-12-07
:sparkles: Compile time flows and CLI simulation update

marcelwa:
## Description

This PR introduces new CMake flags to build individual flows of the *fiction* CLI. Furthermore, it exposes SiDB simulation capabilities via the CLI.

The following CLI commands have been revised:
- `read` now supports the `--sqd`/`-s` flag that reads SiQAD design files as SiDB cell-level layouts
- `print -c` now prints SiDB layouts as H-Si lattices
- `print -c` now prints charge states of SiDB layouts if a CDS is available

The following new CLI commands have been introduced:
- `quicksim` to call the QuickSim algorithm
- `quickexact` to call the QuickExact algorithm
- `temp` to call a critical temperature simulation algorithm based on QuickExact
- `opdom` to call an operational domain computation algorithm and write the resulting domain to a CSV file

## Checklist:


Drewniok:
@marcelwa That's super cool! I will address your points as soon as possible.

include/fiction/algorithms/simulation/sidb/minimum_energy.hpp
```diff
@@ -5,32 +5,56 @@
...
-    static_assert(has_sidb_technology_v<Lyt>, "Lyt is not an SiDB layout");
+    static_assert(std::is_base_of_v<std::input_iterator_tag, typename std::iterator_traits<InputIt>::iterator_category>,
+                  "InputIt must meet the requirements of LegacyInputIterator");
+    static_assert(is_charge_distribution_surface_v<typename std::iterator_traits<InputIt>::value_type>,
+                  "Range must be of charge_distribution_surface objects");
 
-    return std::accumulate(charge_lyts.cbegin(), charge_lyts.cend(), std::numeric_limits<double>::max(),
+    return std::accumulate(first, last, std::numeric_limits<double>::max(),
                            [](const double a, const auto& lyt) { return std::min(a, lyt.get_system_energy()); });
 }
```
marcelwa, line 35: @Drewniok while looking at this function, I'm unsure if this is indeed what you want to compute here. I could be wrong, but don't you want the system energy of the element with the lowest system energy? Would that not be
```c++
return std::min_element(first, last,
                            [](const auto& cds1, const auto& cds2)
                            { return cds1.get_system_energy() < cds2.get_system_energy(); })->get_system_energy();
```

Drewniok:
> @Drewniok In this PR, I've started to expose SiDB simulation via the CLI. You can now use `read --sqd` (or `read -s` for short) SiQAD files and print them via `print -c`. Via commands `quicksim` and `quickexact`, you can generate a CDS, which you can print via `print -c` as well.
> 
> This is pretty basic as far as SiDB simulation on the command line can go, and there are several aspects and details I would like to discuss with you. Here are some things to ponder for you:
> 
>     * What other SiDB algorithms should we expose?
> 
>     * Right now, only the ground state (I think) is saved, how do we deal with the others?
> 
>     * The QuickExact implementation doesn't actually store its global potential variable to the further simulation parameters. I think it should.
> 
>     * QuickExact seems to store the base number additionally, even though it is included in the sim result object via the physical parameters.
> 
>     * Does QuickSim actually respect a given base number? No, right? In that case, the respective option needs to be removed again from the CLI.
> 
> 
> Furthermore TODO:
> 
>     * Add documentation to the CLI chapter in the RTD pages.
> 
> 
> Would be cool if you could play around with it such that we can have a discussion soon.

@marcelwa, I have two questions, comments:

2. Can you explain again or rephrase you question? Are you referring to QuickExact?
4. The base number stored in the additional parameters is the base number actually used for the simulation. This can be different from the base number set by the physical parameters if automatic base number detection is enabled. But I see the confusion! I will change the naming.

include/fiction/algorithms/simulation/sidb/minimum_energy.hpp
Drewniok: @marcelwa You are right. This is quite strange and I think it is a remnant of my first C++ experience ;). I will use ``std::min_element``.

marcelwa:
> > @Drewniok In this PR, I've started to expose SiDB simulation via the CLI. You can now use `read --sqd` (or `read -s` for short) SiQAD files and print them via `print -c`. Via commands `quicksim` and `quickexact`, you can generate a CDS, which you can print via `print -c` as well.
> > This is pretty basic as far as SiDB simulation on the command line can go, and there are several aspects and details I would like to discuss with you. Here are some things to ponder for you:
> > ```
> > * What other SiDB algorithms should we expose?
> > 
> > * Right now, only the ground state (I think) is saved, how do we deal with the others?
> > 
> > * The QuickExact implementation doesn't actually store its global potential variable to the further simulation parameters. I think it should.
> > 
> > * QuickExact seems to store the base number additionally, even though it is included in the sim result object via the physical parameters.
> > 
> > * Does QuickSim actually respect a given base number? No, right? In that case, the respective option needs to be removed again from the CLI.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Furthermore TODO:
> > ```
> > * Add documentation to the CLI chapter in the RTD pages.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would be cool if you could play around with it such that we can have a discussion soon.
> 
> @marcelwa, I have two questions, comments:
> 
> 2. Are you referring to the CLI?
> 3. The base number stored in the additional parameters is the base number actually used for the simulation. This can be different from the base number set by the physical parameters if automatic base number detection is enabled. But I see the confusion! I will change the naming.

My points 1 and 2 referred to the CLI, yes. The others didn't (5 kinda touched upon it).

Thanks for your clarification regarding the base number storage.

cli/cmd/simulation/quickexact.hpp
```diff
@@ -39,6 +39,8 @@ class quickexact_command : public command
         add_option("--mu_minus,-m", physical_params.mu_minus, "Energy transition level (0/-) (unit: eV)", true);
         add_option("--base,-b", physical_params.base,
                    "2-state (neutral/negative) vs. 3-state (positive/neutral/negative) simulation", true);
+        add_option("--base_number_detection, -bd", params.base_number_detection,
```
marcelwa, line 42: @Drewniok shorthand mnemonics can only be single characters and cannot be separated via a space. See the other options for examples.

docs/cli.rst
```diff
@@ -287,7 +287,8 @@ Most important Parameters:
 - Relative permittivity (``-e``).
 - Thomas-Fermi screening (``-l``).
 - Energy transition level (0/-) (``-m``).
-- Base number for the simulation (``-b``)
+- Base number for the simulation (``-b``).
+- Automatic base number detection (``-bd``).
```
marcelwa, line 291: As before

cli/cmd/simulation/quickexact.hpp
marcelwa: Btw, I omitted this option on purpose as I was thinking about omitting `base` as well. It makes sense to always have it set to `ON`, no?

docs/cli.rst
```diff
@@ -215,6 +217,55 @@ throughput (TP) and thereby, the amount of clock cycles the PIs need to be stall
...
+It enumerates all possible charge distributions. However, by incorporating three techniques, namely
+1.) Physically-informed Search Space Pruning, 2.) Partial Solution Caching, and 3.) Effective State Enumeration, it provides
+a significant performance advantage of more than three orders of magnitude over ExGS from SiQAD. For additional details,
+see the `paper <https://www.cda.cit.tum.de/files/eda/2024_aspdac_efficient_exact_simulation.pdf>`_.
+
+Most important Parameters:
+
+- Relative permittivity (``-e``).
+- Thomas-Fermi screening (``-l``).
+- Energy transition level (0/-) (``-m``).
```
marcelwa: Let's provide the variable names here as well, i.e. µ_ etc.

docs/cli.rst
```diff
@@ -215,6 +217,55 @@ throughput (TP) and thereby, the amount of clock cycles the PIs need to be stall
...
+(`max-min diversity distributions <https://onlinelibrary.wiley.com/doi/10.1002/net.20418>`_), are integrated.
+For more in-depth information, refer to the `paper <https://www.cda.cit.tum.de/files/eda/2023_ieeenano_quicksim_physical_simulation.pdf>`_.
+
+Most important Parameters:
+
+- Relative permittivity (``-e``).
+- Thomas-Fermi screening (``-l``).
+- Energy transition level (0/-) (``-m``).
+- Number of iterations (``-i``).
+- Alpha value (``-a``).
```
marcelwa: See above

cli/cmd/simulation/quickexact.hpp
```diff
@@ -0,0 +1,180 @@
...
+                       "specifically for SiDB layouts. It provides a significant performance advantage of more than "
+                       "three orders of magnitude over ExGS from SiQAD.")
+    {
+        add_option("--epsilon_r,-e", physical_params.epsilon_r, "Electric permittivity of the substrate (unit-less)",
+                   true);
+        add_option("--lambda_tf,-l", physical_params.lambda_tf, "Thomas-Fermi screening distance (unit: nm)", true);
+        add_option("--mu_minus,-m", physical_params.mu_minus, "Energy transition level (0/-) (unit: eV)", true);
+        add_option("--base,-b", physical_params.base,
+                   "2-state (neutral/negative) vs. 3-state (positive/neutral/negative) simulation", true);
+        add_option("--base_number_detection, -bd", params.base_number_detection,
```
marcelwa, line 42: You cannot read an `enum class` value from the CLI because there is no known conversion from the stream to this particular `enum class`. You can use a `bool` and then convert it manually. However, as said above, I wouldn't even allow for the option.

cli/cmd/simulation/quickexact.hpp
Drewniok: @marcelwa Yes, sorry! You are right. This option is not that critical for the CLI.

docs/cli.rst
Drewniok: @marcelwa I tried, but I did not manage to write epsilon_r etc. here. Do you have any idea what syntax I can use here?

Drewniok: @marcelwa What do I wrong?

marcelwa: Hmm I don't know for sure, but I think there cannot be a space between the colon and the backtick

Drewniok:
@marcelwa I think all the above TODOs are done. As discussed, it would be convenient to have the critical temperature and part of the operational domain in the CLI. I can support you there!

marcelwa:
> @marcelwa I think all the above TODOs are done. As discussed, it would be convenient to have the critical temperature and part of the operational domain in the CLI. I can support you there!

Perfect, many thanks! 🙏 I'll come up with something and might ask for your feedback once it's done.

marcelwa:
@Drewniok this PR is now feature-complete. If you find some time, I would appreciate your input. It's not urgent.

- The new CLI commands `quickexact`, `quicksim`, `temp`, and `opdom` are available.
- The CLI commands `read` and `print` have been updated to support `read --sqd` and `print -c` for CDS.
- Furthermore, I made some small changes to `critical_temperature` and its tests in order to streamline the code.

include/fiction/algorithms/simulation/sidb/critical_temperature.hpp
```diff
@@ -130,7 +130,7 @@ struct critical_temperature_stats
         if (num_valid_lyt != 0)
         {
             out << fmt::format("'# of physically valid charge configurations': {} | Energy between ground state and "
-                               "first erroneous: {}\n",
+                               "first erroneous in meV: {}\n",
```
marcelwa, line 133: @Drewniok could you also fix this in `cli/commands/temp.hpp`?

Drewniok (approved):
Really cool! Many thanks 🙏

PR #451, opened 2024-06-04
🐛 Multi-output nodes

simon1hofmann:
## Description

Fixes a corner case in `ortho` where a gate has multiple output nodes connected to it.
Thanks to @hibenj for pointing out this solution in his implementation of `orthogonal_ordering_network`.

## Checklist:


hibenj:
The code seems to be working. Before merging you might test the following: I guess you already checked for all benchmarks, that this is working. But you could also check, if the number of added rows corresponds to the number of multi PO nodes. But since you first determine the layout size and then insert the PO nodes, everything should be working perfectly. I just know i had issues with this before.

simon1hofmann:
> The code seems to be working. Before merging you might test the following: I guess you already checked for all benchmarks, that this is working. But you could also check, if the number of added rows corresponds to the number of multi PO nodes. But since you first determine the layout size and then insert the PO nodes, everything should be working perfectly. I just know i had issues with this before.

Additional rows are added based on the number of multi-output nodes in `determine_layout_size()`:

`// for multi-output nodes, add another row
    y += num_multi_output_pos;`

hibenj (commented):
The code seems to be working. But since you already iterate through all pos at the beginning of orthogonal_impl adn therefore compute all multi_out_nodes, you might be able to use them in the main loop of ortho to place the multi_out_nodes. With this implementation you need to iterate through all pos twice.
So i approved it, since its working, but i guess its up to @marcelwa if this implementation makes sense.

hibenj (dismissed):

simon1hofmann:
> The code seems to be working. But since you already iterate through all pos at the beginning of orthogonal_impl and therefore compute all multi_out_nodes, you might be able to use them in the main loop of ortho to place the multi_out_nodes. With this implementation you need to iterate through all pos twice. So i approved it, since its working, but i guess its up to @marcelwa if this implementation makes sense.

The problem ist that multi-output POs usually appear in a different order in `foreach_po` compared to `foreach_node`.
I now added a check for the appearance of multi-output POs in the network and only if they are present, the POs are placed after the main loop. Otherwise, `ortho` places all POs in the main loop.

simon1hofmann:
> > The code seems to be working. But since you already iterate through all pos at the beginning of orthogonal_impl and therefore compute all multi_out_nodes, you might be able to use them in the main loop of ortho to place the multi_out_nodes. With this implementation you need to iterate through all pos twice. So i approved it, since its working, but i guess its up to @marcelwa if this implementation makes sense.
> 
> The problem ist that multi-output POs usually appear in a different order in `foreach_po` compared to `foreach_node`. I now added a check for the appearance of multi-output POs in the network and only if they are present, the POs are placed after the main loop. Otherwise, `ortho` places all POs in the main loop.

Actually, this can also happen for networks without mulit-ouput POs, which explains why the layout for `clpl` was not equivalent. I added another check to test that.

include/fiction/algorithms/physical_design/orthogonal.hpp
```diff
@@ -534,46 +629,56 @@ class orthogonal_impl
...
-                        else
-                        {
-                            const auto anker = layout.create_buf(n_s, po_tile);
+                            tile<Lyt> po_tile{};
+
+                            // determine PO orientation
+                            if (is_eastern_po_orientation_available(ctn, n))
+                            {
+                                po_tile = layout.east(static_cast<tile<Lyt>>(n_s));
+                                ++latest_pos.x;
```
marcelwa, line 652: Instead of manually incrementing x and y positions, rather use `east()` and `south()`

include/fiction/algorithms/physical_design/orthogonal.hpp
```diff
@@ -534,46 +629,56 @@ class orthogonal_impl
...
-                        // determine PO orientation
-                        if (is_eastern_po_orientation_available(ctn, n))
+                        if (place_outputs_last)
                         {
-                            po_tile = layout.east(static_cast<tile<Lyt>>(n_s));
-                            ++latest_pos.x;
+                            if (!is_eastern_po_orientation_available(ctn, n) ||
+                                std::find(multi_out_nodes.begin(), multi_out_nodes.end(), n) != multi_out_nodes.end())
+                            {
+                                ++latest_pos.y;
```
marcelwa, line 639: here too

include/fiction/algorithms/physical_design/orthogonal.hpp
```diff
@@ -534,46 +629,56 @@ class orthogonal_impl
...
+                            // determine PO orientation
+                            if (is_eastern_po_orientation_available(ctn, n))
+                            {
+                                po_tile = layout.east(static_cast<tile<Lyt>>(n_s));
+                                ++latest_pos.x;
+                            }
+                            else
+                            {
+                                po_tile = layout.south(static_cast<tile<Lyt>>(n_s));
+                                ++latest_pos.y;
```
marcelwa, line 657: here too

marcelwa (changes requested):

simon1hofmann:
@marcelwa I refactored the code to always place POs at the end, as this will solve both problems simultaneously.

include/fiction/algorithms/physical_design/orthogonal.hpp
```diff
@@ -379,6 +383,78 @@ mockturtle::signal<Lyt> connect_and_place(Lyt& lyt, const tile<Lyt>& t, const Nt
...
+ * @tparam Ntk Network type that acts as specification.
+ * @param layout Current gate-level layout.
+ * @param ctn Contains the colored network.
+ * @param po_counter Counter for POs.
+ * @param node2pos Mapping from network nodes to layout signals, i.e., a pointer to their position in the layout. The
+ * map is used to fetch location of the fanins. The `mockturtle::node_map` is not updated by this function.
+ */
+template <typename Ntk, typename Lyt>
+void place_outputs(Lyt& layout, const coloring_container<Ntk>& ctn, uint32_t po_counter,
+                   mockturtle::node_map<mockturtle::signal<Lyt>, decltype(ctn.color_ntk)>& node2pos)
```
marcelwa, line 408: If `node2pos` is not updated, it may be `const`

include/fiction/algorithms/physical_design/orthogonal.hpp
```diff
@@ -398,8 +474,26 @@ class orthogonal_impl
 
         mockturtle::node_map<mockturtle::signal<Lyt>, decltype(ctn.color_ntk)> node2pos{ctn.color_ntk};
 
+        // Find multi_output_nodes
```
marcelwa, line 477:
```suggestion
        // find multi-output nodes
```

marcelwa (changes requested):
Just two tiny comments, then it is ready to be merged

PR #456, opened 2024-06-12
✨ Graph-Oriented Layout Design (GOLD)

simon1hofmann:
## Checklist:


include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     */
+    bool high_effort = false;
+    /**
+     * Return first found layout.
+     */
+    bool return_first = false;
+    /**
+     * Timeout limit (in ms).
+     */
+    std::optional<uint64_t> timeout = std::nullopt;
```
marcelwa, line 66: We usually don't use `std::optional` for timeout even though that is a nice idea. Instead, we initialize the value with `std::numeric_limits<uint64_t>::max()`.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     * it sets the timeout to 1000000 milliseconds (1000 seconds) if `high_effort` is `true`,
+     * or to 10000 milliseconds (10 seconds) if `high_effort` is false.
+     *
+     * @param timeout_opt An optional unsigned integer value representing the desired timeout in milliseconds.
+     * @param high_effort A boolean indicating whether a high effort (longer timeout) is desired.
+     */
+    timeout_settings(const std::optional<uint64_t>& timeout_opt, const bool high_effort)
+    {
+        timeout = timeout_opt.value_or(high_effort ? 1000000u : 1000u);
+    }
```
marcelwa, line 277: This is a nice idea. It can however, probably also be implemented directly in the construction of `graph_oriented_layout_design_impl`.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     * it uses that value for the timeout. If `timeout_opt` does not have a value,
+     * it sets the timeout to 1000000 milliseconds (1000 seconds) if `high_effort` is `true`,
+     * or to 10000 milliseconds (10 seconds) if `high_effort` is false.
+     *
+     * @param timeout_opt An optional unsigned integer value representing the desired timeout in milliseconds.
+     * @param high_effort A boolean indicating whether a high effort (longer timeout) is desired.
+     */
+    timeout_settings(const std::optional<uint64_t>& timeout_opt, const bool high_effort)
+    {
+        timeout = timeout_opt.value_or(high_effort ? 1000000u : 1000u);
```
marcelwa, line 276: After some testing, I found 1000 ms as the default runtime value quite insufficient. Even some examples from the TOY folder cannot be generated within 1 s. Ideally, GOLD always produces a result unless a timeout is explicitly given. Where `high_effort_mode` might explore more paths and/or handle more queues additionally. I think "low-effort mode" should at least return something (again, unless a low timeout is explicitly passed).

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+
+/**
+ * Parameters for the graph-oriented layout design algorithm.
+ */
+struct graph_oriented_layout_design_params
+{
+    /**
+     * High effort mode.
+     */
+    bool high_effort = false;
```
marcelwa, line 58: Could you expand on the docstrings here? What does this entail? E.g., increased runtime but might find better result quality layouts, etc. Same for other docstrings.

marcelwa, line 58: `high_effort_mode`?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     */
+    uint64_t best_solution{std::numeric_limits<uint64_t>::max()};
+    /**
+     * Current best solution w.r.t. area after relocating POs.
+     */
+    uint64_t best_optimized_solution{100000};
+    /**
+     * Flag indicating if an initial solution was already found, and that other search space graphs should be pruned.
+     */
+    bool improv_mode{false};
```
marcelwa, line 679: I'm not certain what `improv_mode` stands for. Can we rename this variable to something more telling?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     */
+    uint64_t max_placed_nodes{0ul};
+    /**
+     * Current best solution w.r.t. area.
+     */
+    uint64_t best_solution{std::numeric_limits<uint64_t>::max()};
+    /**
+     * Current best solution w.r.t. area after relocating POs.
+     */
+    uint64_t best_optimized_solution{100000};
```
marcelwa, line 675: Why the default value of `100000` here?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+        Lyt best_lyt{{}, twoddwave_clocking<Lyt>()};
+
+        // Initial setup for networks and nodes to place
+        initialize_networks_and_nodes_to_place();
+
+        // Initial setup for PIs and cost
+        initialize_pis_cost_and_num_expansions();
+
+        // Initial setup for frontiers flags
+        initialize_frontiers_flags();
```
marcelwa, line 564: I'm confused by the fact that we have to call multiple initialize functions (potentially in the exactly correct order) here. That usually indicates that something is wrong with the API. I know it's only an internal API here. But nevertheless, it is error-prone. Can we "hide" all these functions in a single new `initialize` function? Or just merge them into one as some of them are really short?

But also, maybe we should be thinking about the architecture here a little more first. Calling functions that simply change the status of the underlying object, thus "preparing" the state of the object, is a very C-like thing to do. In C++, you would rather have the initialize function return the ready set-up search space graphs that you then use to initialize your member object with.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     * @param dest The destination tile.
+     * @param src_is_new_pos Flag indicating if the source is a new position and should be checked if empty. Defaults to
+     * `false`.
+     * @param dest_is_new_pos Flag indicating if the destination is a new position and should be checked if empty.
+     * Defaults to `false`.
+     * @return A path from `src` to `dest` if one exist, `std::nullopt` otherwise.
+     */
+    [[nodiscard]] std::optional<layout_coordinate_path<ObstrLyt>>
+    check_path(const ObstrLyt& layout, const tile<ObstrLyt>& src, const tile<ObstrLyt>& dest,
+               const bool src_is_new_pos = false, const bool dest_is_new_pos = false) noexcept
```
marcelwa, line 695: There is an ongoing debate in the C++ community about whether to ban `bool` arguments because they make the code so unreadable. On the caller's end, it is simply unclear what `check_path(... , true, false)` means. Plus, it is easy to accidentally swap the parameters. Strongly typed `enum`s are preferrable, i.e. `enum class` for each parameter type. Here, you might even merge them into one, e.g.
```c++
enum class new_po_type
{
NONE,
PI,
PO,
PI_AND_PO
};
```
where these names are to be debated, should just convey the message.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     * @param layout The layout to be checked.
+     * @param src The source tile.
+     * @param dest The destination tile.
+     * @param src_is_new_pos Flag indicating if the source is a new position and should be checked if empty. Defaults to
+     * `false`.
+     * @param dest_is_new_pos Flag indicating if the destination is a new position and should be checked if empty.
+     * Defaults to `false`.
+     * @return A path from `src` to `dest` if one exist, `std::nullopt` otherwise.
+     */
+    [[nodiscard]] std::optional<layout_coordinate_path<ObstrLyt>>
```
marcelwa, line 693: Why the `std::optional`? Paths can be empty.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     * @param nodes_to_place A vector representing the nodes to be placed.
+     * @param current_po The current primary output index.
+     * @param node2pos A dictionary mapping nodes from the network to signals in the layout.
+     * @param pi2node A mapping of primary input nodes to layout nodes.
+     * @return A boolean indicating if a solution was found.
+     */
+    [[nodiscard]] bool place_and_route(const tile<ObstrLyt>& position, ObstrLyt& layout, const tec_nt& network,
+                                       uint64_t& current_node, std::vector<mockturtle::node<tec_nt>>& nodes_to_place,
+                                       uint64_t& current_po, node_dict_type<ObstrLyt, tec_nt>& node2pos,
+                                       mockturtle::node_map<mockturtle::node<ObstrLyt>, tec_nt>& pi2node) noexcept
```
marcelwa, line 1137: Please double-check this function's parameters. We are passing some parameters here that the function can (and will) change. This usually indicates that these parameters should rather be members of some class. Can that be implemented here?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+        layout.obstruct_coordinate({position.x, position.y, 1});
+
+        // Check if solution was found and update max_placed_nodes
+        const auto found_solution = (current_node == nodes_to_place.size());
+        max_placed_nodes          = std::max(current_node, max_placed_nodes);
+
+        return found_solution;
+    }
+    /**
+     * Outputs placement information.
```
marcelwa, line 1194: Could you expand on this docstring?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+     * Computes possible expansions and their priorities for the current vertex in the search space graph.
+     * It handles placement of nodes, checks for valid paths, and finds potential next positions based on priorities.
+     *
+     * @param ssg The search space graph.
+     * @return A pair containing a vector of next positions with their priorities and an optional layout.
+     * If an improved solution is found, the layout is returned.
+     * If the layout is invalid or no improvement is possible, `std::nullopt` is returned.
+     */
+    [[nodiscard]] std::pair<std::vector<std::pair<coord_vec_type<ObstrLyt>, double>>, std::optional<ObstrLyt>>
+    expand(search_space_graph<ObstrLyt>& ssg) noexcept
```
marcelwa, line 1226: This is a really long function. Can it be broken down meaningfully?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+            next_positions.push_back({new_sequence, priority});
+        }
+
+        return {next_positions, std::nullopt};
+    }
+    /**
+     * Initializes the allowed positions for primary inputs (PIs), the cost for each search space graph and the maximum
+     * number of expansions.
+     */
+    void initialize_pis_cost_and_num_expansions() noexcept
```
marcelwa, line 1360: All these initialize functions are quite wild and host a lot of magic numbers. Can we extract constants from them? I.e., `static constexpr uint64_t meaningful_name = 4ul;`

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+    /**
+     * Initializes the enabled search space graphs.
+     */
+    void initialize_frontiers_flags() noexcept
+    {
+        for (uint64_t i = 0ul; i < num_search_space_graphs; ++i)
+        {
+            ssg_vec[i].frontier_flag = true;
+        }
+    }
```
marcelwa, line 1475: Does this have to be a dedicated function?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1523 @@
...
+};
+
+}  // namespace detail
+/**
```
marcelwa, line 1479: Newline before and after namespaces

marcelwa (changes requested):
I've gone over the code in great detail and already made some minor adjustments. Please review them carefully and try to understand why I changed the things I did. But also make sure that I didn't break anything.

The code is already in amazing shape. There are still a few things I would like to point out in this review. I bet we can merge this quickly after some further adjustments, though.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
marcelwa: Can they not even be default-initialized with `true`?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+     */
+    uint64_t current_po;
+    /**
+     * Mapping of nodes to their positions in the layout.
+     */
+    node_dict_type<ObstrLyt, tec_nt>& node2pos;
+    /**
+     * Mapping of primary input nodes to layout nodes.
+     */
+    mockturtle::node_map<mockturtle::node<ObstrLyt>, tec_nt>& pi2node;
```
marcelwa, line 534: Is there a reason for using references here? This is error-prone since the respective variables can run out of scope. Dereferencing (accessing these members) will then cause a `SEGFAULT`. Better remove the `&` here. 

This is assuming that you're passing all instances of the `placement_info` object via reference anyway.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+ * Parameters for the graph-oriented layout design algorithm.
+ */
+struct graph_oriented_layout_design_params
+{
+    /**
+     * In high effort mode, 12 search space graphs are created with varying fanout substitution strategies, allowed PI
+     * placements, and other parameters, compared to only 2 graphs in high efficiency mode. This broader exploration
+     * increases the likelihood of discovering optimal layouts. When a solution is found in any graph, its cost is used
+     * to prune the search in the remaining graphs.
+     */
```
marcelwa, line 60: It should be mentioned that `high_effort_mode` increases the runtime.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+            ssg_vec(num_search_space_graphs)
+    {
+        ntk.substitute_po_signals();
+    }
+    /**
+     * Executes the graph-oriented layout design algorithm and returns the best found layout.
+     *
+     * @return The best layout found by the algorithm.
+     */
+    Lyt run() noexcept
```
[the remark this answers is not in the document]
marcelwa: This then also allows for proper error handling in the `gold` CLI command. See the `exact` command for an example.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+     */
+    bool high_effort_mode = false;
+    /**
+     * Return the first found layout, which might still have a high cost but can be found fast.
+     */
+    bool return_first = false;
+    /**
+     * Timeout limit (in ms).
+     */
+    uint64_t timeout = std::numeric_limits<uint64_t>::max();
```
marcelwa, line 69: When running in high-efficiency mode with no timeout set, GOLD might take forever to complete, even on rather small functions. Maybe it is useful to define a maximum number of paths to evaluate in high-efficiency mode. Without having put much thought into it, maybe we can say explore 1% of paths and then return the best result. While the high-effort mode would evaluate all.

marcelwa: Just an inspiration on how to handle this situation. Your solution may be different of course.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+     * Number of vertices that should be explored from any vertex.
+     */
+    static constexpr uint64_t NUM_VERTEX_EXPANSIONS = 4ul;
+    /**
+     * Checks if there is a path between the source and destination tiles in the given layout.
+     *
+     * @param layout The layout to be checked.
+     * @param src The source tile.
+     * @param dest The destination tile.
+     * @param check_if_empty Enum indicating if the src or dest have to host a new gate and therefore have to be empty.
```
marcelwa, line 706:
```suggestion
     * @param check Enum indicating if the src or dest have to host a new gate and therefore have to be empty.
```

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+     */
+    uint64_t best_optimized_solution{std::numeric_limits<uint64_t>::max()};
+    /**
+     * Flag indicating if an initial solution was already found, and that other search space graphs should be pruned.
+     */
+    bool improve_current_solution{false};
+    /**
+     * Number of vertices that should be explored from any vertex.
+     */
+    static constexpr uint64_t NUM_VERTEX_EXPANSIONS = 4ul;
```
marcelwa, line 699: Should this be a parameter?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+        /* mark children */
+        this->foreach_fanout(n, [this](const auto& fo) { create_topo_rec(fo); });
+    }
+
+    std::vector<node> topo_order;
+};
+/**
+ * Enum to specify which tiles should be checked if they are empty.
+ */
+enum class check_if_empty
```
marcelwa, line 496: I'm not super happy with the name and docstring of this `enum class`. I can infer from neither what this is used for. Can we find a more descriptive naming and documentation here?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+
+        const auto min_x = std::max(pre1_t.x, pre2_t.x) + (pre1_t.x == pre2_t.x ? 1 : 0);
+        const auto min_y = std::max(pre1_t.y, pre2_t.y) + (pre1_t.y == pre2_t.y ? 1 : 0);
+
+        // check if path from previous tiles to new tile and from new tile to drain exist
+        auto check_tile = [&](uint64_t x, uint64_t y)
+        {
+            const tile<ObstrLyt> new_pos{min_x + x, min_y + y, 0};
+
+            const auto path_1 = check_path(layout, pre1_t, new_pos, check_if_empty::DEST);
```
marcelwa, line 926: If there is no `path_2`, we might as well rename this variable to `path` or `p`.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+            graph.num_expansions                    = NUM_VERTEX_EXPANSIONS;
+        }
+    }
+    /**
+     * Initializes the networks and nodes to place.
+     */
+    void initialize_networks_and_nodes_to_place() noexcept
+    {
+        // prepare nodes to place for each topology view
+        auto prepare_nodes_to_place = [](auto& network, auto& nodes_to_place)
```
marcelwa, line 1437:
```suggestion
        const auto prepare_nodes_to_place = [](auto& network, auto& nodes_to_place) noexcept
```

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1574 @@
...
+                                                                  {true, true},
+                                                                  {true, false},
+                                                                  {false, true},
+                                                                  {true, true},
+                                                                  {true, false},
+                                                                  {false, true},
+                                                                  {true, true},
+                                                                  {true, false},
+                                                                  {false, true},
+                                                                  {true, true}}};
```
marcelwa, line 1416: As mentioned above, this can probably be made a little more meaningful with `enum class` objects

cli/cmd/physical_design/gold.hpp
```diff
@@ -0,0 +1,117 @@
...
+            env->out() << "[w] no logic network in store" << std::endl;
+            ps = {};
+            return;
+        }
+
+        if (is_set("timeout"))
+        {
+            // convert timeout entered in seconds to milliseconds
+            ps.timeout *= 1000;
+        }
```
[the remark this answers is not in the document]
marcelwa: Could be that that was by design. The Mugen library handling the one-pass synthesis maybe receives its timeout in seconds. Please double-check that.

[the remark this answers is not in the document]
simon1hofmann: You're right, `mugen` receives the timeout in seconds already.
```
/**
 * Sets a timeout in seconds for the solving process, where 0 allows for unlimited time.
 */
uint32_t timeout = 0u;
```

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
simon1hofmann: I have another proposal:

In high-effiency mode, the timeout is set to 10s.
After the 10s, the best found layout is returned, but if no solution was found, the timeout is ignored until the first solution is found.
This behavior is of course disable when the user sets a timeout.

marcelwa: Sounds good. Let's try this! :muscle:

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1626 @@
...
+    }
+    /**
+     * Initializes the layout with minimum width
+     *
+     * @param min_layout_width The minimum width of the layout.
+     * @return The initialized layout.
+     */
+    ObstrLyt initialize_layout(uint64_t min_layout_width)
+    {
+        Lyt lyt{{min_layout_width - 1, 1 - 1, 1}, twoddwave_clocking<Lyt>()};
```
marcelwa, line 1294:
```suggestion
        Lyt lyt{{min_layout_width - 1, 0, 1}, twoddwave_clocking<Lyt>()};
```

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1626 @@
...
+        // if no timeout was set and high-effort mode is disabled, set timeout to 10s
+        if (!timeout_set && !ps.high_effort_mode)
+        {
+            timeout = 10000u;
+        }
+        bool timeout_limit_reached = false;
+        // main A* loop
+        while (!timeout_limit_reached)
+        {
+            for (auto& ssg : ssg_vec)
```
marcelwa, line 625: Would you like to add progress bar support like in `ortho` and `exact`?

simon1hofmann: Good idea!

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1646 @@
...
+        if (!timeout_set && !ps.high_effort_mode)
+        {
+            timeout = 10000u;
+        }
+        bool timeout_limit_reached = false;
+
+#if (PROGRESS_BARS)
+        // initialize a progress bar
+        mockturtle::progress_bar bar{ssg_vec[0].network.size(), "[i] arranging layout: |{0}|"};
+        uint64_t                 max_placed_nodes_tracker = 0u;
```
marcelwa, line 630: There are some problems with the progress bar as of right now. It immediately displays a completed bar.

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1646 @@
...
+                    }
+                    else
+                    {
+                        ssg.frontier_flag = false;
+                    }
+
+#if (PROGRESS_BARS)
+                    if (max_placed_nodes > max_placed_nodes_tracker)
+                    {
+                        max_placed_nodes_tracker = max_placed_nodes;
```
marcelwa, line 685: Is this the intended behavior? I expected the tracker to be slowly increased with every step in the graph that has been evaluated.

cli/cmd/physical_design/gold.hpp
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * Logs the resulting information in a log file.
+     *
+     * @return JSON object containing information about the physical design process.
+     */
+    nlohmann::json log() const override
+    {
+        return nlohmann::json{{"runtime in seconds", mockturtle::to_seconds(st.time_total)}};
```
marcelwa, line 81: Is the runtime the only thing you want to log?

cli/cmd/physical_design/gold.hpp
```diff
@@ -0,0 +1,135 @@
...
+    void execute() override
+    {
+        // error case: empty logic network store
+        if (store<fiction::logic_network_t>().empty())
+        {
+            env->out() << "[w] no logic network in store" << std::endl;
+            ps = {};
+            return;
+        }
+
```
marcelwa, line 62: Maybe add another check warning if `ps.num_vertex_expansions == 0`?

include/fiction/algorithms/physical_design/graph_oriented_layout_design.hpp
```diff
@@ -0,0 +1,1646 @@
...
+     * @param p The parameters for the graph-enhanced layout search algorithm.
+     * @param st The statistics object to record execution details.
+     */
+    graph_oriented_layout_design_impl(const Ntk& src, const graph_oriented_layout_design_params& p,
+                                      graph_oriented_layout_design_stats& st) :
+            ntk{convert_network<tec_nt>(src)},
+            ps{p},
+            pst{st},
+            timeout{ps.timeout},
+            num_vertex_expansions{ps.num_vertex_expansions},
```
marcelwa, line 594: We don't need this variable anymore. We can just access the parameters.

test/algorithms/physical_design/graph_oriented_layout_design.cpp
```diff
@@ -0,0 +1,187 @@
...
+    check_eq(ntk, *layout4);
+
+    // Timeout limit reached
+    params.timeout     = 0;
+    const auto layout6 = graph_oriented_layout_design<gate_layout>(ntk, params, &stats);
+
+    CHECK(!layout6.has_value());
+}
+
+TEST_CASE("Name conservation after graph-enhanced layout search", "[graph_oriented_layout_design]")
```
marcelwa, line 149:
```suggestion
TEST_CASE("Name conservation after graph-oriented layout design", "[graph_oriented_layout_design]")
```

test/algorithms/physical_design/graph_oriented_layout_design.cpp
```diff
@@ -0,0 +1,187 @@
...
+    check_graph_oriented_layout_design_equiv<Lyt>(
+        blueprints::fanout_substitution_corner_case_network<technology_network>());
+    check_graph_oriented_layout_design_equiv<Lyt>(blueprints::inverter_network<technology_network>());
+    check_graph_oriented_layout_design_equiv<Lyt>(blueprints::clpl<technology_network>());
+    check_graph_oriented_layout_design_equiv<Lyt>(
+        blueprints::one_to_five_path_difference_network<technology_network>());
+    check_graph_oriented_layout_design_equiv<Lyt>(blueprints::nand_xnor_network<technology_network>());
+}
+
+TEST_CASE("Layout equivalence", "[algorithms]")
```
marcelwa, line 64:
```suggestion
TEST_CASE("Layout equivalence after graph-oriented layout design", "[graph-oriented-layout-design]")
```

test/algorithms/physical_design/graph_oriented_layout_design.cpp
```diff
@@ -0,0 +1,187 @@
...
+{
+    SECTION("Cartesian layouts")
+    {
+        using gate_layout = gate_level_layout<clocked_layout<tile_based_layout<cartesian_layout<offset::ucoord_t>>>>;
+
+        check_graph_oriented_layout_design_equiv_all<gate_layout>();
+    }
+}
+
+TEST_CASE("Gate library application", "[graph_oriented_layout_design]")
```
marcelwa, line 74: Replace the tag with `graph-oriented-layout-design` (hyphens instead of underscores) everywhere. Also make sure that the test names are unique across the entire test suite. Otherwise, it could lead to issues on some systems assuming that test names are overloaded.

test/benchmark/graph_oriented_layout_design.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <bill/sat/interface/common.hpp>
+#include <mockturtle/networks/aig.hpp>
+
+using namespace fiction;
+
+TEST_CASE("Benchmark Graph-Oriented Layout Design", "[benchmark]")
+{
+    using gate_layout = gate_level_layout<clocked_layout<tile_based_layout<cartesian_layout<offset::ucoord_t>>>>;
+
+    auto ntk    = blueprints::mux21_network<mockturtle::aig_network>();
```
marcelwa, line 27: Shall we use a slightly more complex circuit than a 2:1 MUX for benchmarking?

experiments/graph_oriented_layout_design/graph_oriented_layout_design.cpp
```diff
@@ -0,0 +1,100 @@
+//
+// Created by Simon Hofmann on 30.01.24.
+//
+#include "fiction_experiments.hpp"
+
+#include <fiction/algorithms/physical_design/graph_oriented_layout_design.hpp>  // graph enhanced layout search algorithm
```
marcelwa, line 6:
```suggestion
#include <fiction/algorithms/physical_design/graph_oriented_layout_design.hpp>  // graph-oriented layout design algorithm
```

marcelwa (changes requested):
Many thanks for your hard work on this PR :pray: We are almost there. I just spotted a few issues and minor inconsistencies. After they have been addressed, it is finally ready for merging :)

marcelwa (approved):
Many, many thanks for your hard work on this algorithm and the PR! :pray:

PR #462, opened 2024-06-20
:green_heart: Fix compilation of Python wheels for Windows

Drewniok:
## Description

Github has released an update to MSVC v144 for the ``windows-latest`` runner. This could cause this problem as discussed [here](https://github.com/pypa/cibuildwheel/issues/1875). So we have to use ``windows-2019`` until this is fixed.

## Checklist:

### DIFF
--- a/.github/workflows/pyfiction-pypi-deployment.yml
+++ b/.github/workflows/pyfiction-pypi-deployment.yml
@@ -50,10 +50,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - { os: ubuntu-latest, arch: x64 }
-          - { os: macos-latest, arch: x64 }
-          - { os: macos-latest, arch: arm64 }
-          - { os: windows-2019, arch: x64}
+          - { os: ubuntu-22.04, arch: x64 }
+          - { os: macos-13, arch: x64 }
+          - { os: macos-14, arch: arm64 }
+          - { os: windows-2019, arch: x64 }
         python: ['cp38', 'cp39', 'cp310', 'cp311', 'cp312']
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/pyfiction-pypi-deployment.yml
+++ b/.github/workflows/pyfiction-pypi-deployment.yml
@@ -53,7 +53,7 @@ jobs:
           - { os: ubuntu-latest, arch: x64 }
           - { os: macos-latest, arch: x64 }
           - { os: macos-latest, arch: arm64 }
-          - { os: windows-2022, arch: x64}
+          - { os: windows-2019, arch: x64}
         python: ['cp38', 'cp39', 'cp310', 'cp311', 'cp312']
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/pyfiction-pypi-deployment.yml
+++ b/.github/workflows/pyfiction-pypi-deployment.yml
@@ -53,7 +53,7 @@ jobs:
           - { os: ubuntu-latest, arch: x64 }
           - { os: macos-latest, arch: x64 }
           - { os: macos-latest, arch: arm64 }
-          - { os: windows-2012, arch: x64, toolset: v143 }
+          - { os: windows-2022, arch: x64, toolset: v143 }
         python: ['cp38', 'cp39', 'cp310', 'cp311', 'cp312']
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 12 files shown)


Drewniok:
@marcelwa After spending several hours, I can now say that the problem is caused by multithreading in _QuickSim_. I have tried to solve the problem by using ``std::future`` without success. 
The problem is solved as soon as multithreading is not used. I think this is quite an important issue and topic since multithreading is important for our tools (_QuickSim_, _OpDomain_, etc.).

marcelwa:
I agree. What exactly is it about the multi-threading that causes this problem do you think?

marcelwa:
Have you tried activating the respective sanitizers?

.github/workflows/python-bindings.yml
```diff
@@ -69,7 +75,7 @@ jobs:
       - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1.2
         with:
-          key: '${{matrix.os}}-pyfiction'
+          key: '${{matrix.os}}-${{matrix.toolset}}-pyfiction'
```
marcelwa, line 78: I don't think you need the toolset key because there is only one build per operating system

.github/workflows/pyfiction-pypi-deployment.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           - { os: ubuntu-latest, arch: x64 }
           - { os: macos-latest, arch: x64 }
           - { os: macos-latest, arch: arm64 }
-          - { os: windows-latest, arch: x64 }
+          - { os: windows-2019, arch: x64}
```
marcelwa, line 56:
```suggestion
          - { os: windows-2019, arch: x64 }
```

.github/workflows/pyfiction-pypi-deployment.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           - { os: ubuntu-latest, arch: x64 }
           - { os: macos-latest, arch: x64 }
           - { os: macos-latest, arch: arm64 }
```
marcelwa, line 55: Is there a reason for not specifying the exact versions here, too?

bindings/pyfiction/include/pyfiction/algorithms/simulation/sidb/quicksim.hpp
```diff
@@ -44,9 +44,7 @@ inline void quicksim(pybind11::module& m)
                        DOC(fiction_quicksim_params_iteration_steps))
         .def_readwrite("alpha", &fiction::quicksim_params::alpha, DOC(fiction_quicksim_params_alpha))
         .def_readwrite("number_threads", &fiction::quicksim_params::number_threads,
-                       DOC(fiction_quicksim_params_number_threads))
-
-        ;
+                       DOC(fiction_quicksim_params_number_threads));
```
marcelwa, line 47: Please revert this change
```suggestion
                       DOC(fiction_quicksim_params_number_threads))
                       
                       ;
```

test/benchmark/simulation.cpp
```diff
@@ -61,7 +62,7 @@ TEST_CASE("Benchmark simulators", "[benchmark]")
 
     BENCHMARK("QuickExact")
     {
-        const quickexact_params<lattice> sim_params{sidb_simulation_parameters{2, -0.32}};
+        const quickexact_params<cell<lattice>> sim_params{sidb_simulation_parameters{2, -0.32}};
```
marcelwa, line 65: If this change was necessary, how did it compile before? This looks like something where only one of the solutions would compile.

marcelwa (changes requested):

test/benchmark/simulation.cpp
Drewniok: Yes, the reason is that the benchmark code is not compiled for the CI. Should we change that?

.github/workflows/pyfiction-pypi-deployment.yml
Drewniok: No, I will update this part.

test/benchmark/simulation.cpp
marcelwa: Good point. If we do we need to exclude its execution in the test run. Otherwise, this would just consume too much time. I'll take a note of this. Later problem 😄

Drewniok: Yes, that's the point. Then, we just leave it for now. Agreed! 😁

PR #466, opened 2024-06-28
:construction_worker: Updated the Ubuntu CI to include 24.04 and newer compiler versions

marcelwa:
## Description

Dropped `ubuntu-20.04` and added `ubuntu-24.04`. Additionally, updated the C++ compilers to newer versions

## Checklist:

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -43,7 +43,11 @@ jobs:
         exclude:
           - os: ubuntu-24.04
             compiler: clang++12
+          - os: ubuntu-24.04
+            compiler: clang++13
         include:
+          - os: ubuntu-22.04
+            compiler: clang++15
           - os: ubuntu-22.04
             compiler: g++-12
             cppstandard: -DCMAKE_CXX_STANDARD=20
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -38,24 +38,11 @@ jobs:
   build_and_test:
     strategy:
       matrix:
-        os: [ ubuntu-20.04, ubuntu-22.04 ]
-        compiler: [ g++-10, g++-11, g++-12, clang++-10, clang++-11, clang++-12 ]
-        exclude:
-          - os: ubuntu-20.04
-            compiler: g++-11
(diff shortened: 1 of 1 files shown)


marcelwa:
@Drewniok the build with GCC 14 fails here because of the `convert_to_siqad_coordinates` function. It is a GCC bug. Do you think, you could rewrite the function anyway to reduce nesting and return type inference to assist the compiler?

marcelwa:
> @Drewniok the build with GCC 14 fails here because of the `convert_to_siqad_coordinates` function. It is a GCC bug. Do you think, you could rewrite the function anyway to reduce nesting and return type inference to assist the compiler?

The issue occurs on macOS as well. See, e.g., here: https://github.com/cda-tum/fiction/actions/runs/10387750687/job/28761888544?pr=499

Hope that helps with reproducibility.

Drewniok:
> @Drewniok the build with GCC 14 fails here because of the `convert_to_siqad_coordinates` function. It is a GCC bug. Do you think, you could rewrite the function anyway to reduce nesting and return type inference to assist the compiler?
are al
@marcelwa Oh, that's not good. Sure, I'll have a look asap. I'll try my best, but this convert function has already caused me a lot of worries.

marcelwa:
> > @Drewniok the build with GCC 14 fails here because of the `convert_to_siqad_coordinates` function. It is a GCC bug. Do you think, you could rewrite the function anyway to reduce nesting and return type inference to assist the compiler?
> > are al
> > @marcelwa Oh, that's not good. Sure, I'll have a look asap. I'll try my best, but this convert function has already caused me a lot of worries.

Understandable. It is not a simple function. I spent some time myself (maybe 30 min) and could not fix the issue. It probably requires dissecting the code a bit more.

Drewniok:
@marcelwa Finally I could fix it! I am not sure where to put the code change. We can close #501 and change the title of #499 to make it clear that the code was also slightly changed. What do you think?

marcelwa:
> @marcelwa Finally I could fix it! I am not sure where to put the code change. We can close #501 and change the title of #499 to make it clear that the code was also slightly changed. What do you think?

Agreed. That would be good. Thank you so much for investigating the problem.

PR #486, opened 2024-07-29
👷 Disabled running computationally intensive unit tests via the CI in Debug build mode

Drewniok:
## Description

This PR moves all computationally intensive unit tests into a separate test suite called ``quality``. These tests will only run in release mode within the CI pipeline, accelerating the workflow (> 20 % runtime improvement for macos CI).

## Checklist:


marcelwa:
@Drewniok running expensive tests in `Release` mode only, is a worthwhile addition. Thanks a lot! I don't understand why they had to be moved to a separate file. By assigning them an additional tag `[quality]` on top of their original tag, we can achieve the same result I think. This way, everything can stay in place and thus logically sorted.

Drewniok:
> @Drewniok running expensive tests in `Release` mode only, is a worthwhile addition. Thanks a lot! I don't understand why they had to be moved to a separate file. By assigning them an additional tag `[quality]` on top of their original tag, we can achieve the same result I think. This way, everything can stay in place and thus logically sorted.

That's a good point! I think, I just did it like in ``mockturtle``. But I agree, it's better to remove the separate file again. Thanks!

marcelwa (approved):
Many thanks! :pray:

PR #487, opened 2024-07-30
🐛 Use correct label for --exclude-regex

Drewniok:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Checklist:

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -138,7 +138,7 @@ jobs:
 
       - name: Test (Debug)
         working-directory: ${{github.workspace}}/build_debug
-        run: ctest -C Debug --verbose --output-on-failure --repeat until-pass:3 --parallel 4 --label-exclude "quality"
+        run: ctest -C Debug --verbose --output-on-failure --repeat until-pass:3 --parallel 4 -LE "quality"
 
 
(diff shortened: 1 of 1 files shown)


marcelwa:
> @marcelwa Unfortunately, I have not found a way to get it to work. Maybe you have an idea how to fix the problem. Maybe some additional lines are needed in the corresponding cmake. Thank you for your help.

I'll have a look. Thanks for trying!

marcelwa:
@Drewniok I looked into the problem. As of right now, Catch2 doesn't pass the tag labels on to `ctest`. This is revealed when executing `ctest --print-labels` which gives an empty output (`No Labels Exist`). On the other hand, the `-E` parameter works to exclude named test files (not what we want). 

The issue was raised in the Catch2 tracker 5 years ago already: https://github.com/catchorg/Catch2/issues/1590 Hence, I doubt that it will be fixed. 

Thereby indeed your initial solution of moving all tests to an extra file and then excluding it via `-E <filename>` could be a viable workaround if this is desperately needed.

Drewniok:
@marcelwa Thanks for checking! That's a pity!

PR #538, opened 2024-10-10
✨ Timeout for `post_layout_optimization` and `wiring_reduction`

simon1hofmann:
## Description

Added a timeout parameter to `post_layout_optimization` and `wiring_reduction`. 
If a timeout is set for `post_layout_optimization`, the timeout for the call of `wiring_reduction` inside `post_layout_optimization` is handled automatically.

Additionally, `post_layout_optimization` was refactored for readability and a benchmark script was added.

## Checklist:


include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -441,12 +469,17 @@ layout_coordinate_path<Lyt> get_path_and_obstruct(Lyt& lyt, const tile<Lyt>& sta
...
+ * @param timeout_limit_reached Reference to a boolean flag that is set to `true` if the timeout is reached.
  * @param planar_optimization Only allow relocation if a crossing-free wiring can be found. Defaults to false.
  * @return `true` if the gate was moved successfully, `false` otherwise.
  */
 template <typename Lyt>
 bool improve_gate_location(Lyt& lyt, const tile<Lyt>& old_pos, const tile<Lyt>& max_non_po,
-                           const uint64_t max_gate_relocations, const bool planar_optimization = false) noexcept
+                           const uint64_t                                       max_gate_relocations,
+                           const std::chrono::high_resolution_clock::time_point start, const uint64_t timeout,
+                           bool& timeout_limit_reached, const bool planar_optimization = false) noexcept
```
marcelwa, line 482: I would prevent passing timeout parameters around and instead make them private members of the `post_layout_optimization_impl` class.

Particularly here, you have to pass three different variables for timeout tracking. That's excessive and makes it easy to use this function wrong.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -427,6 +436,25 @@ layout_coordinate_path<Lyt> get_path_and_obstruct(Lyt& lyt, const tile<Lyt>& sta
...
+ * if the timeout is exceeded, and returns the remaining time.
+ *
+ * @param start The start time in milliseconds.
+ * @param timeout The timeout limit in milliseconds.
+ * @param timeout_limit_reached Reference to a boolean flag that is set to `true` if the timeout is reached.
+ *
+ * @return Remaining time in milliseconds before timeout, or `0` if timeout has been reached.
+ */
+inline uint64_t update_timeout(const std::chrono::high_resolution_clock::time_point start, const uint64_t timeout,
+                               bool& timeout_limit_reached) noexcept
```
marcelwa, line 450: Same here. The parameters should actually be members.

include/fiction/algorithms/physical_design/wiring_reduction.hpp
```diff
@@ -1044,70 +1064,114 @@ class wiring_reduction_impl
...
+        // create an obstruction layout based on the original layout
         auto layout = obstruction_layout<Lyt>(plyt);
 
+        // initialize the list of wires to delete
         layout_coordinate_path<wiring_reduction_layout_type<coordinate<Lyt>>> to_delete = {};
 
         bool found_wires = true;
 
+        // lambda to update the timeout status and calculate remaining time
+        auto update_timeout = [&]() -> void
```
marcelwa, line 1081: Instead of a default capture (`[&]`), specify each name to capture separately. That prevents accidental copy where reference was intended and vice versa.

marcelwa (changes requested):
That's a great addition, thanks a lot!

simon1hofmann:
> That's a great addition, thanks a lot!

Thanks a lot for the review, I took the opportunity to refactor the algorithm a little bit to make it more readable.

test/benchmark/post_layout_optimization.cpp
```diff
@@ -0,0 +1,65 @@
...
+    {
+        post_layout_optimization<gate_layout>(layout.clone());
+    };
+
+    params.max_gate_relocations = 0;
+
+    BENCHMARK("post_layout_optimization: wiring reduction only")
+    {
+        post_layout_optimization<gate_layout>(layout.clone(), params);
+    };
```
marcelwa, line 40: Currently, the `clone()` call is also part of the runtime measurement. Is that intended behavior?

marcelwa, line 40: Currently, only one call receives parameters. This might be confusing and is prone to breaking should default parameters change in the future. Can we create two separate parameter objects with explicitly set defaults?

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -918,65 +364,92 @@ class post_layout_optimization_impl
...
         bool reduced_wiring          = true;
+        timeout_limit_reached        = false;
 
         if (max_gate_relocations == 0)
         {
-            fiction::wiring_reduction_stats wiring_reduction_stats{};
-            fiction::wiring_reduction(layout, &wiring_reduction_stats);
+            // perform wiring reduction only when no gate relocations are allowed
+            fiction::wiring_reduction_stats  wiring_reduction_stats{};
+            fiction::wiring_reduction_params wiring_reduction_params{};
```
marcelwa, line 400: You might want to set explicit defaults here as well.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -918,65 +364,92 @@ class post_layout_optimization_impl
...
-            while (moved_at_least_one_gate || reduced_wiring)
+            // iteratively optimize the layout until no more improvements can be made or timeout is reached
+            while ((moved_at_least_one_gate || reduced_wiring) && !timeout_limit_reached)
             {
                 reduced_wiring = false;
-                fiction::wiring_reduction_stats wiring_reduction_stats{};
-                if (moved_at_least_one_gate && !ps.optimize_pos_only)
+
+                // update the remaining timeout
+                fiction::wiring_reduction_params wiring_reduction_params{};
```
marcelwa, line 418: See above.

Also, why the duplication? Can we make these params a member of the class?

marcelwa: Same with the statistics then.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -918,65 +364,92 @@ class post_layout_optimization_impl
...
+                    if (wiring_reduction_stats.area_improvement != 0ull ||
+                        wiring_reduction_stats.wiring_improvement != 0ull)
                     {
                         reduced_wiring = true;
                     }
                 }
 
-                std::vector<tile<Lyt>> gate_tiles{};
+                // gather all relevant gate tiles for relocation
+                std::vector<tile<Lyt>> gate_tiles;
```
marcelwa, line 435:
```suggestion
                std::vector<tile<Lyt>> gate_tiles{};
```

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -918,65 +364,92 @@ class post_layout_optimization_impl
...
                 gate_tiles.reserve(layout.num_gates() + layout.num_pis() - layout.num_pos());
                 layout.foreach_node(
                     [&layout, &gate_tiles](const auto& node)
                     {
-                        if (const tile<Lyt> tile = layout.get_tile(node);
-                            (layout.is_gate(node) && !layout.is_wire(node)) || layout.is_fanout(node) ||
+                        const tile<Lyt> tile = layout.get_tile(node);
```
marcelwa, line 440: Variable name `tile` collides with type `tile<Lyt>`.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -918,65 +364,92 @@ class post_layout_optimization_impl
...
                         reduced_wiring = true;
                     }
                 }
 
-                std::vector<tile<Lyt>> gate_tiles{};
+                // gather all relevant gate tiles for relocation
+                std::vector<tile<Lyt>> gate_tiles;
                 gate_tiles.reserve(layout.num_gates() + layout.num_pis() - layout.num_pos());
                 layout.foreach_node(
                     [&layout, &gate_tiles](const auto& node)
```
marcelwa, line 438:
```suggestion
                    [&layout, &gate_tiles](const auto& node) noexcept
```

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -986,43 +459,66 @@ class post_layout_optimization_impl
                         max_non_po.y = std::max(max_non_po.y, gate_tile.y);
                     }
                 }
+
+                // reset the gate movement flag
                 moved_at_least_one_gate = false;
+
+                // attempt to relocate each gate tile
                 for (auto& gate_tile : gate_tiles)
```
marcelwa, line 467: Can `gate_tile` be `const`?
```suggestion
                for (const auto& gate_tile : gate_tiles)
```

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
marcelwa, line 440: It doesn't make a difference here, but it is generally preferred to move the condition variable declaration inside the `if` statement condition to limit its scope. This leads to less pollution and faster resource deallocation.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -986,43 +459,66 @@ class post_layout_optimization_impl
...
                 {
-                    if (!ps.optimize_pos_only || (ps.optimize_pos_only && layout.is_po_tile(gate_tile)))
+                    if (!timeout_limit_reached)
                     {
-                        if (detail::improve_gate_location(layout, gate_tile, max_non_po, max_gate_relocations,
-                                                          ps.planar_optimization))
+                        bool should_optimize =
+                            !ps.optimize_pos_only || (ps.optimize_pos_only && layout.is_po_tile(gate_tile));
+                        if (should_optimize)
```
marcelwa, line 473: Looks like `should_optimize` can be `const` and moved into the `if` condition.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+     */
+    tile<Lyt> max_non_po{0, 0};
+    /**
+     * Timeout limit reached.
+     */
+    bool timeout_limit_reached = false;
+    /**
+     * Start time.
+     */
+    std::chrono::time_point<std::chrono::high_resolution_clock> start;
```
marcelwa, line 549: It is a best practice to declare members in the same order they are constructed in, putting all the ones instantiated by the constructor first. In this case, you would move `start` over `max_gate_relocations`.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+    /**
+     * Utility function to move wires that cross over empty tiles down one layer. This can happen if the wiring of a
+     * gate is deleted.
+     *
+     * @param lyt Gate-level layout.
```
marcelwa, line 554: It's an `ObstrLyt`

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+     * @param deleted_coords Tiles that got deleted.
+     */
+    void fix_wires(ObstrLyt& lyt, const std::vector<tile<ObstrLyt>>& deleted_coords) noexcept
+    {
+        std::unordered_set<tile<ObstrLyt>> moved_tiles{};
```
marcelwa, line 559: You could try to use `phmap::parallel_flat_hash_set` here instead of `std::unordered_set` to see if you get a performance improvement. Analogously for your other usages of associative STL containers.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+     * @param end_tile The ending coordinate of the path.
+     * @return The computed path as a sequence of coordinates in the layout.
+     */
+    layout_coordinate_path<ObstrLyt> get_path_and_obstruct(ObstrLyt& lyt, const tile<ObstrLyt>& start_tile,
+                                                           const tile<ObstrLyt>& end_tile)
+    {
+        using dist = twoddwave_distance_functor<ObstrLyt, uint64_t>;
+        using cost = unit_cost_functor<ObstrLyt, uint8_t>;
+        static a_star_params params{};
+        params.crossings = !ps.planar_optimization;
```
marcelwa, line 764: Even though `params` is `static` here, this assignment is executed every time the function is called. Maybe you want to move it inside `params`' constructor.

marcelwa: And also maybe you want to rename `params` to state that they are related to A*.

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+
+        layout_coordinate_path<ObstrLyt> path =
```
marcelwa, line 766:
```suggestion
        const auto path =
```

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+     * - determining coordinates that would improve the layout
+     * - testing all those coordinates by moving the gate to each one and checking if a new wiring can be found
+     * - if a new coordinate is found and wiring is possible, it is applied and incoming signals are updated
+     * - if no better coordinate is found, the old wiring is restored
+     *
+     * @param lyt 2DDWave-clocked cartesian obstruction gate-level layout.
+     * @param old_pos Old position of the gate to be moved.
+     * @return `true` if the gate was moved successfully, `false` otherwise.
+     */
+    bool improve_gate_location(ObstrLyt& lyt, const tile<ObstrLyt>& old_pos) noexcept
```
marcelwa, line 806: This is a really long function. Can it be broken down meaningfully?

include/fiction/algorithms/physical_design/post_layout_optimization.hpp
```diff
@@ -1035,6 +531,567 @@ class post_layout_optimization_impl
...
+
+                lyt.move_node(lyt.get_node(fanout), fanout, fout_signals);
+            }
+            return false;
+        }
+
+        return true;
+    }
 };
 }  // namespace detail
```
marcelwa, line 1096:
```suggestion
};

}  // namespace detail
```

marcelwa (changes requested):
Totally agree with the restructuring. It made me aware of a few things that can be improved here and there. Please have a look.

test/benchmark/post_layout_optimization.cpp
simon1hofmann: As `post_layout_optimization` optimizes the layout directly, I need to pass the unoptimized layout. Furthermore, the computational overhead of cloning the layout should be negligible and also a constant time overhead.

include/fiction/algorithms/physical_design/wiring_reduction.hpp
```diff
@@ -1136,10 +1213,11 @@ class wiring_reduction_impl
  *
  * @tparam Lyt Cartesian gate-level layout type.
  * @param lyt The 2DDWave-clocked layout whose wiring is to be reduced.
- * @param pst Pointer to a `wiring_reduction_stats` object to record runtime statistics.
+ * @param ps Parameters.
+ * @param pst Statistics.
  */
 template <typename Lyt>
-void wiring_reduction(const Lyt& lyt, wiring_reduction_stats* pst = nullptr) noexcept
+void wiring_reduction(const Lyt& lyt, wiring_reduction_params ps = {}, wiring_reduction_stats* pst = nullptr) noexcept
```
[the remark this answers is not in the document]
simon1hofmann: Right now, it is not really consistent in the other algorithms.
Should we aim for:
`void wiring_reduction(const Lyt& lyt, wiring_reduction_params params = {}, wiring_reduction_stats* stats = nullptr)` 

for the main function and the for the `_impl_ classes:

```wiring_reduction_impl(const Lyt& lyt, const wiring_reduction_params& ps, wiring_reduction_stats& pst) :
            plyt{lyt},
            params{p},
            stats{st}}```
            
 or:

```wiring_reduction_impl(const Lyt& lyt, const wiring_reduction_params& params, wiring_reduction_stats& stats) :
            plyt{lyt},
            ps{params},
            pst{stats}}```

I can then take care of this for the whole repository in another PR.

File: include/fiction/algorithms/physical_design/post_layout_optimization.hpp (hunk 15 above)
Comment (marcelwa) [reply include/fiction/algorithms/physical_design/post_layout_optimization.hpp:764]: Ah, of course. Because `ps.planar_optimization` is only known at call time. However, `a_star_params` is already instantiated when the program starts.

Btw, avoid naming collision. The variable has the same name as its type right now.

File: test/benchmark/post_layout_optimization.cpp (hunk 4 above)
Comment (marcelwa) [reply test/benchmark/post_layout_optimization.cpp:40]: You could clone outside of the `BENCHMARK` call. It should be negligible, but up to you.

File: include/fiction/algorithms/physical_design/wiring_reduction.hpp (hunk 19 above)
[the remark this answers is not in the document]
Comment (marcelwa) [reply include/fiction/algorithms/physical_design/wiring_reduction.hpp:1220]: The first solution:

```c++
wiring_reduction_impl(const Lyt& lyt, const wiring_reduction_params& ps, wiring_reduction_stats& pst) :
            plyt{lyt},
            params{p},
            stats{st}}
```

Many thanks! :pray:

test/benchmark/post_layout_optimization.cpp
simon1hofmann: That leads to the same problem, as then the cloned layout is optimized directly.

PR #555, opened 2024-10-31
📥 Added SiQAD SiDB gate library 

Drewniok:
## Description

Since a significant amount of the experiments use the SiQAD SiDB gate library (https://ieeexplore.ieee.org/document/8963859), I propose to add it. This PR includes it with an additional tag for the canvas SiDBs (``cell_type::LOGIC``).

## Checklist:


marcelwa:
Many thanks! I'm just a bit confused by the folder structure. If I were looking for the AND gate, I would expect the path to be `sidb_gate_libraries/siqad/and.sqd`. Could you explain your design decision?

Drewniok:
> Many thanks! I'm just a bit confused by the folder structure. If I were looking for the AND gate, I would expect the path to be `sidb_gate_libraries/siqad/and.sqd`. Could you explain your design decision?

Hmm, we could live without the subfolders. This structure is still from the time when we did not have the ``bdl_input_iterator``. So up to 5 files were stored for each gate. 
However, at the moment it is quite convenient for the experiment because the name of the gate is always given by the folder and the file itself can be named arbitrarily.

Drewniok:
> > > Many thanks! I'm just a bit confused by the folder structure. If I were looking for the AND gate, I would expect the path to be `sidb_gate_libraries/siqad/and.sqd`. Could you explain your design decision?
> > 
> > 
> > Hmm, we could live without the subfolders. This structure is still from the time when we did not have the `bdl_input_iterator`. So up to 5 files were stored for each gate. However, at the moment it is quite convenient for the experiment because the name of the gate is always given by the folder and the file itself can be named arbitrarily.
> 
> I understand where this is coming from. Regarding the convenience, you could just as easily argue the opposite: The name of the gate is given by the file name and you could put whatever files in the folder and they would automatically work with the experiments if they simply iterate over all SQD files in the `sidb_gate_libraries/siqad/` directory.

Yes, I know, it is true 😅. I will change it for the Bestagon and SiQAD gates, right?

marcelwa:
> > > > Many thanks! I'm just a bit confused by the folder structure. If I were looking for the AND gate, I would expect the path to be `sidb_gate_libraries/siqad/and.sqd`. Could you explain your design decision?
> > > 
> > > 
> > > Hmm, we could live without the subfolders. This structure is still from the time when we did not have the `bdl_input_iterator`. So up to 5 files were stored for each gate. However, at the moment it is quite convenient for the experiment because the name of the gate is always given by the folder and the file itself can be named arbitrarily.
> > 
> > 
> > I understand where this is coming from. Regarding the convenience, you could just as easily argue the opposite: The name of the gate is given by the file name and you could put whatever files in the folder and they would automatically work with the experiments if they simply iterate over all SQD files in the `sidb_gate_libraries/siqad/` directory.
> 
> Yes, I know, it is true 😅. I will change it for the Bestagon and SiQAD gates, right?

Many thanks :pray:

marcelwa (approved):

PR #574, opened 2024-11-13
✅ Update Unit Tests and Exclude Long-Running Tests in Debug Mode Using ``NDEBUG``

Drewniok:
## Description

This PR updates unit tests and uses ``NDEBUG`` to exclude long-running tests in debug mode.

## Checklist:


marcelwa:
This should be possible without CMake changes. The vast majority of compilers set a flag called `NDEBUG` when in Release mode (not debug). So you can check for
```c++
#ifndef NDEBUG
    // Debug mode
#else
    // Release mode
#endif
```
While not technically guaranteed by ISO C++, the standard itself depends on it, too, as it uses it for conditionally disabling `assert` statements: https://en.cppreference.com/w/c/error/assert

It is thus fairly safe to use as all compilers we continuously test for should implement `NDEBUG`.

Drewniok:
> This should be possible without CMake changes. The vast majority of compilers set a flag called `NDEBUG` when in Release mode (not debug). So you can check for
> 
> ```c++
> #ifndef NDEBUG
>     // Debug mode
> #else
>     // Release mode
> #endif
> ```
> 
> While not technically guaranteed by ISO C++, the standard itself depends on it, too, as it uses it for conditionally disabling `assert` statements: https://en.cppreference.com/w/c/error/assert
> 
> It is thus fairly safe to use as all compilers we continuously test for should implement `NDEBUG`.

Wow! Thank you!

Drewniok:
> Your changes are fine and we can gladly merge them like this. However, if you are feeling fancy, you could define new macros `RELEASE_ONLY_TEST_CASE`, `RELEASE_ONLY_TEMPLATE_TEST_CASE`, and `RELEASE_ONLY_SECTION` in a separate file that wrap the `NDEBUG` checks for us.

That's a great idea. Let's merge this and we can add it later.

marcelwa (approved):

marcelwa:
Don't forget to update the PR description given the latest changes ;)

PR #600, opened 2024-12-01
✨ Added rectangular skeletons (16.896 nm x 16.896 nm) with design script.

Drewniok:
## Description

This PR introduces rectangular skeletons measuring ``16.896 nm x 16.896 nm`` for designing gate implementations. Moreover, it adds a script to design a corresponding gate library.

Some results:

``CX (input pattern: 01 left/top):``

``AND (input pattern: 01 left/top):``

``Fan-out (input pattern: 1 at the top):``

## Checklist:


docs/changelog.rst
```diff
@@ -11,6 +11,7 @@ Unreleased
 Added
 #####
 - Experiments:
+    - Added rectangular skeletons and a script for designing a gate library.
```
marcelwa, line 14:
```suggestion
    - Added rectangular SiDB gate skeletons and a script for designing a corresponding library
```

experiments/quickcell/quickcell_rectangular_gate_library.cpp
```diff
@@ -0,0 +1,175 @@
...
+    const auto rectangular_1i_left_1o_right =
+        read_sqd_layout<sidb_100_cell_clk_lyt_siqad>(fmt::format("{}/{}", folder, "rectangular_1i_left_1o_right.sqd"));
+
+    const auto rectangular_1i_top_2o_right_down = read_sqd_layout<sidb_100_cell_clk_lyt_siqad>(
+        fmt::format("{}/{}", folder, "rectangular_1i_top_2o_right_down.sqd"));
+
+    const auto rectangular_1i_top_2o_left_right = read_sqd_layout<sidb_100_cell_clk_lyt_siqad>(
+        fmt::format("{}/{}", folder, "rectangular_1i_top_2o_left_right.sqd"));
+
+    const auto num_canvas_sidbs = 3;
```
marcelwa, line 92:
```suggestion
    constexpr auto num_canvas_sidbs = 3u;
```

experiments/quickcell/quickcell_rectangular_gate_library.cpp
```diff
@@ -0,0 +1,175 @@
...
+        design_sidb_gates_stats                  stats_quickcell{};
+
+        if (gate_name == "cx_2i_top_left_2o_down_right" || gate_name == "ha_2i_top_left_2o_down_right" ||
+            gate_name == "hourglass_2i_top_left_2o_down_right")
+        {
+            params.number_of_sidbs = 4;
+            params.canvas          = {{17, 8, 0}, {27, 14, 0}};
+            quickcell_design =
+                design_sidb_gates(rectangular_2i_top_left_2o_down_right, truth_table, params, &stats_quickcell);
+            params.number_of_sidbs = num_canvas_sidbs;
```
marcelwa, line 113: Instead of resetting the number of SiDBs here, I would explicitly set them below. Do avoid code duplication, you could create one big `else` in which you set the value once, and have several `if ... else if` inside.

experiments/quickcell/quickcell_rectangular_gate_library.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    for (const auto& [truth_table, gate_name] : truth_tables_and_names)
+    {
+        std::vector<sidb_100_cell_clk_lyt_siqad> quickcell_design{};
+        design_sidb_gates_stats                  stats_quickcell{};
+
+        if (gate_name == "cx_2i_top_left_2o_down_right" || gate_name == "ha_2i_top_left_2o_down_right" ||
+            gate_name == "hourglass_2i_top_left_2o_down_right")
+        {
+            params.number_of_sidbs = 4;
```
marcelwa, line 109: Why not a constant for this value as well?

experiments/quickcell/quickcell_rectangular_gate_library.cpp
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        else
+        {
+            quickcell_design =
+                design_sidb_gates(rectangular_2i_top_left_1o_right, truth_table, params, &stats_quickcell);
+        }
+
+        // Write the layout to a file
+        // write_sqd_layout(quickcell_design.front(), fmt::format("{}/{}", folder, gate_name + ".sqd"));
```
marcelwa, line 150: Why is this commented out?

marcelwa (changes requested):

experiments/quickcell/quickcell_rectangular_gate_library.cpp
Drewniok: sure, we can also activate it.

Drewniok:
@marcelwa, can you tell me if anything else needs to be done before this is merged?

marcelwa:
> @marcelwa, can you tell me if anything else needs to be done before this is merged? 

Did you not want to integrate your new library as an `fcn_gate_library` subclass to map to when using `apply_gate_library`?

marcelwa:
> > > @marcelwa, can you tell me if anything else needs to be done before this is merged?
> > 
> > Did you not want to integrate your new library as an `fcn_gate_library` subclass to map to when using `apply_gate_library`?
> 
> I would rather do this in a separate PR. The reason for this is that this PR is more the basis for further exploration in this direction. If we decide to go in that direction, we can gladly integrate into the `fcn_gate_library`. But for now, I would rather use it to experiment with single-gate implementations.

That's alright. Makes sense to me.

marcelwa (approved):

PR #604, opened 2024-12-04
:snake: Add a `.pyi` file for integration with type checking, documentation, and IDE auto-completion

marcelwa:
## Description

This PR adds a `.pyi` file that defines a type interface for all *pyfiction* classes and functions. This helps with integration with type checking, documentation, and IDE auto-completion.

## Checklist:


marcelwa:
Thank you for this — the need it identified is real and still open, and it is now tracked
as #1099.

Closing here because the approach has been overtaken. This pull request maintains the
stub by hand, and `mnt.pyfiction` exposes roughly 370 symbols, so the file has to be
updated by hand for every binding change with nothing to detect when it drifts. The
bindings have since moved to nanobind, which ships a stub generator, so #1099 generates
the stubs instead:

- a `nox -s stubs` session running the generator and formatting the output;
- the generated `.pyi` files committed, so editors get them without a build;
- a pattern file for the signatures the generator cannot infer;
- a CI job that regenerates and fails if the result differs from what is committed.

That last point is what this pull request could not offer and what makes the stubs stay
correct.

#1099 is blocked on #1096 and #1097, which restructure `include/fiction`'s directories and
introduce nested namespaces. Generating stubs before that lands would mean regenerating
them immediately afterwards.

Drafted with AI assistance; the symbol count and the current absence of any `py.typed` or
`.pyi` in the repository were verified against the working tree.

PR #624, opened 2025-01-07
💚 Updated CI to support macOS 15

Drewniok:
## Description

This pull request includes updates to the GitHub actions workflow for MacOS builds. 

Even though the MacOS 15 GitHub runners are still in beta, I think there are at least two reasons to update our workflow:
1.) The MacOS 15 runners have been available for almost 3 months. So we can assume that the major bugs have been fixed and the release to GA should happen soon.
2.) MacOS 15 is already used by many people (MacOS 15.2 Sequoia) is already available. So I would argue that we should support it in _fiction_.

## Checklist:


Drewniok:
@marcelwa, please feel free to let me know what you think of this test matrix.

.github/workflows/python-bindings.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ubuntu-latest, macos-13, macos-14, windows-latest]
+        runs-on: [ubuntu-latest, macos-13, macos-14, macos-15, windows-latest]
```
marcelwa, line 49: Why use `macos-15` here but `macos-15-xlarge` above?

marcelwa: We only need one Intel and one ARM macOS runner here

Drewniok: > We only need one Intel and one ARM macOS runner here

so if I choose ``macos-15-xlarge``, it is okay, right? 

macos-15-xlarge: arm
macos-14: x64

marcelwa: In that case, `macos-13` is superfluous. Again, we only need one Intel and one ARM runner here for macOS.

Drewniok: ah, I see! Now, I know what you mean. Thanks!

marcelwa: I just double-checked with GitHub's documentation. `macos-13` is the only tag that has Intel runners available. `macos-14` and `macos-15` are M1-based, i.e., ARM. The `xlarge` runners we cannot use because they are only available for Enterprise customers.

Drewniok: Oh, thank you for double-checking.

.github/workflows/macos.yml
```diff
@@ -35,34 +35,35 @@ env:
   Z3_VERSION: 4.13.0
 
 jobs:
-  build_and_test:
+  build_and_test_other_macos:
```
marcelwa, line 38: Can this change be reverted?

.github/workflows/macos.yml
```diff
@@ -75,7 +76,6 @@ jobs:
         with:
           python-version: "3.13.x"
 
-      # Setup TBB for parallel STL algorithms via Homebrew
```
marcelwa, line 78: Can we please leave the comment in?

.github/workflows/macos.yml
```diff
@@ -104,9 +104,8 @@ jobs:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
 
       # Build and test pipeline for Debug mode
-
       - name: Create Build Environment (Debug)
-        run: cmake -E make_directory ${{github.workspace}}/build_debug
```
marcelwa, line 109: While your change is equivalent, I'd prefer the explicit workspace reference

.github/workflows/macos.yml
```diff
@@ -131,12 +130,11 @@ jobs:
 
       - name: Test (Debug)
         working-directory: ${{github.workspace}}/build_debug
-        run: ctest -C Debug --verbose --output-on-failure --repeat until-pass:3 --parallel 4 --exclude-regex "quality"
+        run: ctest -C Debug --verbose --output-on-failure --repeat until-pass:3 --parallel 4
 
       # Build and test pipeline for Release mode
-
       - name: Create Build Environment (Release)
-        run: cmake -E make_directory ${{github.workspace}}/build_release
```
marcelwa, line 139: Same here

.github/workflows/macos.yml
```diff
@@ -131,12 +130,11 @@ jobs:
 
       - name: Test (Debug)
         working-directory: ${{github.workspace}}/build_debug
-        run: ctest -C Debug --verbose --output-on-failure --repeat until-pass:3 --parallel 4 --exclude-regex "quality"
```
marcelwa, line 134: Good catch

.github/workflows/macos.yml
```diff
@@ -35,34 +35,35 @@ env:
...
-            compiler: g++-14
+            architecture: x64
+          - os: macos-15
+            architecture: x64
+          - os: macos-13
+            architecture: arm64
 
-    name: 🍎 ${{matrix.os}} with ${{matrix.compiler}}
-    runs-on: ${{matrix.os}}
+    name: 🍎 ${{ matrix.os }} with ${{ matrix.compiler }} with ${{ matrix.architecture }}
```
marcelwa, line 65: The added spaces here pose an inconsistency, or? I thought `pre-commit` would automatically address this 🤔

marcelwa (changes requested):

marcelwa:
@Drewniok I don't think `g++` is able to compile code for Apple Silicon chips. We can safely exclude all `g++` versions from macOS 14 and 15.

Drewniok:
> @Drewniok I don't think `g++` is able to compile code for Apple Silicon chips. We can safely exclude all `g++` versions from macOS 14 and 15.

okay, thanks!

.github/workflows/macos.yml
Drewniok: Yes, weird! I also assumed that ``pre-commit`` would fix it.

.github/workflows/macos.yml
```diff
@@ -38,30 +38,43 @@ jobs:
...
+        os: [macos-13, macos-14, macos-15]
+        architecture: [x64, arm64]
+        compiler: [g++-13, g++-14, clang++]
         include:
           - os: macos-13
             architecture: x64
           - os: macos-14
             architecture: arm64
+          - os: macos-15
+            architecture: arm64
```
marcelwa, line 50: Now that you explicitly include the architectures in the matrix above, these statements are redundant since all architectures are matched against all operating systems by default. It is then left to exclude the combinations you don't want. I'd prefer the other way around as it was before because we only need these three includes then and no excludes for the architectures.

marcelwa (changes requested):

.github/workflows/macos.yml
Drewniok: ah, great! Thanks!

PR #646, opened 2025-01-25
✨ Added temperature-aware operational domain simulation.

Drewniok:
## Description

This PR expands the operational domain simulation to include temperature. It is now possible to simulate temperature at each parameter point, creating a temperature-aware operational domain.

## Checklist:


include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -178,28 +179,73 @@ struct operational_domain
      * different physical parameters, the parameters are stored with the corresponding operating status.
      */
     locked_parallel_flat_hash_map<Key, Value> operational_values{};
+    /**
+     * This can store different information depending on the use case. If the critical temperature is simulated on top
+     * of the operational domain, it is stored here.
+     */
+    locked_parallel_flat_hash_map<Key, MetricType> metric_values{};
```
marcelwa, line 186: Why is it necessary to distinguish between a metric and a value? Couldn't a temperature domain just be an `operational_domain<parameter_point, double>` where `Value` is `double` and represents the critical temperature? Maybe we'd find a better name for `Value` then :wink:

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -178,28 +179,73 @@ struct operational_domain
...
+    /**
+     * Adds a value to the metric values.
+     *
+     * @param key The key to add the value to.
+     * @param value The value to add.
+     */
+    void add_metric_value(const Key& key, const MetricType& value)
+    {
+        metric_values.try_emplace(key, value);
+    }
```
marcelwa, line 230:
```suggestion
    /**
     * Adds a metric value to the operational domain.
     *
     * @param key The key to add the value to.
     * @param value The value to add.
     */
    void add_metric_value(const Key& key, const MetricType& value)
    {
        metric_values.try_emplace(key, value);
    }
```

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -178,28 +179,73 @@ struct operational_domain
...
+     * This can store different information depending on the use case. If the critical temperature is simulated on top
+     * of the operational domain, it is stored here.
+     */
+    locked_parallel_flat_hash_map<Key, MetricType> metric_values{};
     /**
      * This function retrieves the value associated with the provided key from the operational domain. If
      * the key is found in the domain, its corresponding value is returned. Otherwise, `std::nullopt`
      * is returned.
      *
      * @param key The key to look up.
```
marcelwa, line 192:
```suggestion
     * @param key The key to look up in the operational domain.
```

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -178,28 +179,73 @@ struct operational_domain
...
+    }
+    /**
+     * Returns the operational domain.
+     *
+     * @return The operational domain.
+     */
+    [[nodiscard]] locked_parallel_flat_hash_map<Key, Value> get_operational_domain() const
     {
         return operational_values;
     }
```
marcelwa, line 239: This function might be confusing since the associated class is called an `operational_domain` already.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -178,28 +179,73 @@ struct operational_domain
...
+    /**
+     * Returns the metric values. If no metric values are present, `std::nullopt` is returned.
+     *
+     * @return The metric values.
+     */
+    [[nodiscard]] locked_parallel_flat_hash_map<Key, Value> get_metric_values() const
+    {
+        return metric_values;
+    }
```
marcelwa, line 248:
```suggestion
    /**
     * Returns the metric values.
     *
     * @return The metric values.
     */
    [[nodiscard]] locked_parallel_flat_hash_map<Key, Value> get_metric_values() const
    {
        return metric_values;
    }
```

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -229,6 +275,21 @@ struct operational_domain_value_range
  */
 struct operational_domain_params
 {
+    /**
+     * This enumeration defines whether specific metrics, such as critical temperature,
+     * are simulated within the operational domain.
+     */
+    enum class metric_simulation : uint8_t
```
marcelwa, line 282: Actually, this enumeration doesn't define such behavior. It simply offers the options to do so. The respective logic is handled elsewhere.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -229,6 +275,21 @@ struct operational_domain_value_range
...
+    {
+        /**
+         * Simulates the critical temperature on top of the operational domain.
+         */
+        CRITICAL_TEMPERATURE_SIM,
+        /**
+         * Disables metric simulation, meaning no metrics are simulated within the operational domain.
+         */
+        DISABLED
+    };
```
marcelwa, line 292: In unison with my comments from above regarding operational status vs. metric, we can use this enumeration to toggle between operational status, critical temperature, ... for computation. Currently, `DISABLED` just doesn't want to make sense to me.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: I'd suggest `get_operational_values` in accordance with the function below.

include/fiction/io/write_operational_domain.hpp
```diff
@@ -103,14 +103,17 @@ namespace detail
  * The operational status is a binary value represented by specified tags in `params` indicating whether the simulation
  * parameters are within the operational domain or not.
  *
+ * @tparam OpDomain The type of the operational domain.
  * @param opdom The operational domain to be written. It contains a mapping from sets of simulation parameters
- * (represented as a pair of sweep parameters for the X and Y dimensions) to their operational status.
+ * (represented as a pair of sweep parameters for the X and Y dimensions) to their operational status, and it can
+ include metric values (e.g., the critical temperature on top of that).
```
marcelwa, line 109:
```suggestion
 * (represented as a pair of sweep parameters for the X, Y, and Z dimensions) to their operational status, and it can
 include metric values (e.g., the critical temperature on top of that).
```

marcelwa, line 109: What does it mean that it can "include" metric values? How are those represented? Can we get an example file excerpt here?

include/fiction/io/write_operational_domain.hpp
```diff
@@ -121,45 +124,94 @@ inline void write_operational_domain(const operational_domain<parameter_point, o
         throw std::invalid_argument("unsupported number of dimensions in the given operational domain");
     }
 
-    if (num_dimensions == 1)
+    if (!opdom.metric_values.empty())
```
marcelwa, line 127: If possible, I would always start with a non-negated case in a condition. It makes it easier to read for humans, particularly when it comes to the `else` case where we'd otherwise have to think about a double negation.

include/fiction/io/write_operational_domain.hpp
```diff
@@ -121,45 +124,94 @@ inline void write_operational_domain(const operational_domain<parameter_point, o
...
-            continue;
+            writer.write_line(detail::sweep_parameter_to_string(opdom.dimensions[0]),
+                              detail::sweep_parameter_to_string(opdom.dimensions[1]), "operational status", "ct");
         }
+        else  // num_dimensions == 3
+        {
+            writer.write_line(detail::sweep_parameter_to_string(opdom.dimensions[0]),
+                              detail::sweep_parameter_to_string(opdom.dimensions[1]),
+                              detail::sweep_parameter_to_string(opdom.dimensions[2]), "operational status", "ct");
+        }
```
marcelwa, line 143: `"ct"` is hard-coded here. Would be cool to have a mapping from metrics to strings where this symbol is fetched. Also, I would spell it out as `"critical temperature"` since it will occur only once in the header of the file

include/fiction/io/write_operational_domain.hpp
```diff
@@ -180,15 +232,17 @@ inline void write_operational_domain(const operational_domain<parameter_point, o
  * The operational status is a binary value represented by specified tags in `params` indicating whether the simulation
  * parameters are within the operational domain or not.
  *
- * @param opdom The operational domain to be written. It contains a mapping from sets of simulation parameters
- * (represented as a pair of sweep parameters for the X and Y dimensions) to their operational status.
+* @tparam OpDomain The type of the operational domain.
+* @param opdom The operational domain to be written. It contains a mapping from sets of simulation parameters
+* (represented as a pair of sweep parameters for the X and Y dimensions) to their operational status, and it can include
+metric values (e.g., the critical temperature on top of that).
```
marcelwa, line 238: See above

test/algorithms/simulation/sidb/operational_domain.cpp
```diff
@@ -41,17 +42,30 @@ void check_op_domain_params_and_operational_status(
             const auto& sweep_param = params.sweep_dimensions[d];
             const auto& coord_value = coord.parameters[d];
 
-            CHECK(sweep_param.min <= coord_value);
-            CHECK(sweep_param.max >= coord_value);
+            CHECK(sweep_param.min <= (coord_value + physical_constants::POP_STABILITY_ERR));
+            CHECK(sweep_param.max >= (coord_value - physical_constants::POP_STABILITY_ERR));
```
marcelwa, line 46: Not necessarily a comment regarding this PR specifically, but I noticed we're using `POP_STABILITY_ERR` a lot in places where we actually mean an *epsilon*. Do you think, we should keep an eye on this? Should it be renamed? Its usage restricted? Is it a physical constant even as indicated by its namespace?

test/algorithms/simulation/sidb/operational_domain.cpp
```diff
@@ -1358,3 +1383,80 @@ TEMPLATE_TEST_CASE("Grid search to determine the operational domain. The operati
         }
     }
 }
+
+TEST_CASE("Bestagon AND gate operational domain and temperature computation, using cube coordinates",
+          "[operational-domain]")
```
marcelwa, line 1388: We say `cube` coordinates here but it looks like SiQAD coordinates are used

marcelwa (changes requested):
Thanks a lot for the addition! I like the content a lot, and I bet we'll get on the same page regarding its implementation in no time. Let's gladly chat about it soon :slightly_smiling_face:

Drewniok:
> Thanks a lot for the addition! I like the content a lot, and I bet we'll get on the same page regarding its implementation in no time. Let's gladly chat about it soon 🙂

Many thanks for your review! 🙏

test/algorithms/simulation/sidb/operational_domain.cpp
Drewniok: I completely agree, and I have thought about the same thing a few times! It is basically an epsilon that is bigger than the one from ``std::limits``. Do you have any idea how we could call it? Should we rename the namespace to ``constants`` if we decide to keep it there?

include/fiction/io/write_operational_domain.hpp
Drewniok: True! Thanks!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: Hmm, I think I know what you mean. But I also find it confusing. Right now I think of the code like this:
We compute the operational domain, and the operational state is always computed, right? So, in my opinion, there is no need to specify if it is computed. Now, with the new feature, we have the ability to add other metrics on top of the operational state. If this is desired, we specify which metric with the enumeration, if no metric is desired on top of the operational domain, we disable this option, which is the default. Is that also your understanding of the current implementation. 
My concern is that if you as a user decide to select ``critical temperature`` instead of ``operational state``, you might think that only the ``ct`` is computed and not the operational state. But in fact both are there.

Drewniok: I also wonder how to deal with multiple metrics. Let's say we add _defect clearance_ and _BBR_ in the future. Would it still be nice to add all 6 combinations to the enum class. Do you know a best practice for this scenario?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: True! There is no need for that.

test/algorithms/simulation/sidb/operational_domain.cpp
marcelwa: Multiple ideas come to mind here. It could make sense to move it into its own `namespace constants`, leaving `namespace physical_constants` intact. Alternatively, we use nested namespaces, e.g.,
```c++
namespace constants
{

// here goes new epsilon (and possibly other non-physical constants like pi)

namespace physical
{

// here go physical constants

}

}
```
where current physical constants would then be accessed via `constants::physical::<name_of_constant>`.

Regarding the implementation of that new epsilon itself, it would be cool to base it off C++'s epsilon, something like
```c++
inline constexpr double scaled_epsilon = std::numeric_limits<double>::epsilon() * 10;
```

We might want to workshop the naming, though. It could make sense to name it following its usage, e.g., `precision_threshold`, `error_margin`, or `numerical_tolerance`.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
marcelwa: There may be another discussion about code cleanliness vs. efficiency. Ideally, we can achieve both even if it takes more work. Let's see.

The first question would be: is it useful to store the operational value **and** the critical temperature in a single operational domain? My understanding is that if a parameter point is non-operational, there will be no critical temperature. Is that correct? If so, we may as well store the critical temperature exclusively. Following this logic, we need only one additional template type, namely the one defining the mapped type of the operational domain. Thus, only one map. 

I would like to hear your thoughts before going too deep into hypothetical terrain here :slightly_smiling_face:

test/algorithms/simulation/sidb/operational_domain.cpp
Drewniok: Great ideas! I love them! Many thanks!

test/algorithms/simulation/sidb/operational_domain.cpp
Drewniok: Good catch!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -579,7 +515,7 @@ class operational_domain_impl
         // add the neighbors of each operational point to the queue
         for (const auto& [param_point, status] : op_domain.get_domain())
         {
-            if (status == operational_status::OPERATIONAL)
+            if (std::get<0>(status) == operational_status::OPERATIONAL)
```
marcelwa, line 518: This only works if the first entry in the tuple of mapped values in `OpDomain` is of type `operational_status`. This is not guaranteed if you're calling this function, or? As per usual, there are multiple possible solutions to this: We could for example restrict calls via `static_assert`s.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -721,17 +656,19 @@ class operational_domain_impl
      * @param lyt SiDB cell-level layout that is simulated and compared to the given CDS.
      * @return All physically valid physical parameters and the excited state number.
      */
-    [[nodiscard]] operational_domain<parameter_point, uint64_t>
-    grid_search_for_physically_valid_parameters(Lyt& lyt) noexcept
+    template <typename OpDomain_target>
+    [[nodiscard]] OpDomain_target grid_search_for_physically_valid_parameters(Lyt& lyt) noexcept
```
marcelwa, line 660: Invalid case style for a template type

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -770,11 +707,11 @@ class operational_domain_impl
...
         sidb_simulation_parameters simulation_parameters = params.operational_params.simulation_parameters;
 
-        for (const auto& [param_point, status] : op_domain.operational_values)
+        for (const auto& [param_point, status] : op_domain.get_domain())
         {
-            if constexpr (std::is_same_v<OpDomain, operational_domain<parameter_point, operational_status>>)
+            if constexpr (std::is_same_v<OpDomain, operational_domain>)
             {
-                if (status == operational_status::NON_OPERATIONAL)
+                if (std::get<0>(status) == operational_status::NON_OPERATIONAL)
```
marcelwa, line 714: See above

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -1054,9 +991,9 @@ class operational_domain_impl
      */
     [[nodiscard]] inline std::optional<operational_status> has_already_been_sampled(const step_point& sp) const noexcept
     {
-        if (const auto v = contains_key(op_domain.operational_values, to_parameter_point(sp)); v.has_value())
+        if (const auto v = contains_key(op_domain.get_domain(), to_parameter_point(sp)); v.has_value())
         {
-            return v.value();
+            return std::get<0>(v.value());
         }
```
marcelwa, line 997: See above

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -1054,9 +991,9 @@ class operational_domain_impl
      */
     [[nodiscard]] inline std::optional<operational_status> has_already_been_sampled(const step_point& sp) const noexcept
     {
```
marcelwa, line 993: The multiple distinctions between different operational domain types inside the function feel kinda clunky, no? I think this can be improved by providing dedicated implementations of `has_already_been_sampled` for each operational domain type supported via `std::enable_if`.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -1617,21 +1582,23 @@ class operational_domain_impl
...
+OpDomain operational_domain_grid_search(const Lyt& lyt, const std::vector<TT>& spec,
+                                        const operational_domain_params& params = {},
+                                        operational_domain_stats*        stats  = nullptr)
 {
     static_assert(is_cell_level_layout_v<Lyt>, "Lyt is not a cell-level layout");
     static_assert(has_sidb_technology_v<Lyt>, "Lyt is not an SiDB layout");
     static_assert(kitty::is_truth_table<TT>::value, "TT is not a truth table");
+    static_assert(std::is_same_v<OpDomain, operational_domain> ||
+                      std::is_same_v<OpDomain, temperature_operational_domain>,
+                  "OpDomain must be operational_domain or temperature_operational_domain");
```
marcelwa, line 1595: I would have kept these functions as is and provided additional ones called `temperature_domain_grid_search`, `temperature_domain_random_sampling`, etc. That simplifies the caller interface and makes it harder for users to use them wrong.

marcelwa (changes requested):
Great to see how quickly you implemented the results of our discussion! In this review, I focused in-depth on the new class hierarchy of the SiDB domains. You probably agree that it is most important to get these classes straight before moving on to the depending functions. I have some thoughts that hopefully help us make the new types even more intuitive to use. If anything is unclear, let's gladly discuss.

include/fiction/algorithms/simulation/sidb/defect_influence.hpp
```diff
@@ -98,13 +98,8 @@ enum class defect_influence_status : uint8_t
...
  */
 template <typename Lyt>
-struct defect_influence_domain
-{
-    /**
-     * This stores for each defect position the condition of the layout.
-     */
-    locked_parallel_flat_hash_map<typename Lyt::cell, defect_influence_status> influence_information{};
-};
+using defect_influence_domain = sidb_simulation_domain<typename Lyt::cell, defect_influence_status>;
```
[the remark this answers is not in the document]
Drewniok: Alright! Good point!

Drewniok:
> Great to see how quickly you implemented the results of our discussion! In this review, I focused in-depth on the new class hierarchy of the SiDB domains. You probably agree that it is most important to get these classes straight before moving on to the depending functions. I have some thoughts that hopefully help us make the new types even more intuitive to use. If anything is unclear, let's gladly discuss.

Many thanks for the quick feedback! Much appreciated!

docs/technology/simulation.rst
```diff
@@ -88,7 +88,7 @@ Physical Constants
 .. doxygenvariable:: EPSILON
 .. doxygenvariable:: ELEMENTARY_CHARGE
 .. doxygenvariable:: K_E
-.. doxygenvariable:: POP_STABILITY_ERR
+.. doxygenvariable:: SCALED_EPSILON
```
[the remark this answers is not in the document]
Drewniok: Hmm, that's not great! I have already realized that there was some problem with the renaming. I will check again for misnames.

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,102 @@
...
+ * multithreaded environments.
+ *
+ * @tparam Key The type of the key used to identify entries in the domain.
+ * @tparam Value1 The first value type stored in the tuple associated with each key.
+ * @tparam Valuetypes Additional value types stored in the tuple.
+ */
+template <typename Key, typename Value1, typename... Valuetypes>
+struct sidb_simulation_domain
+{
+    locked_parallel_flat_hash_map<Key, std::tuple<Value1, Valuetypes...>> domain_values{};
```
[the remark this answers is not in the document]
Drewniok: Agreed! Let's make it to a class.

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,102 @@
...
+ * associated with values stored as tuples. It uses a `locked_parallel_flat_hash_map` to ensure
+ * thread-safe access to the stored data. This is especially useful for parallel simulations or
+ * multithreaded environments.
+ *
+ * @tparam Key The type of the key used to identify entries in the domain.
+ * @tparam Value1 The first value type stored in the tuple associated with each key.
+ * @tparam Valuetypes Additional value types stored in the tuple.
+ */
+template <typename Key, typename Value1, typename... Valuetypes>
+struct sidb_simulation_domain
```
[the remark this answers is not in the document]
Drewniok: Thank you!

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,102 @@
...
+typename MapType::const_iterator find_key_with_tolerance(const MapType& map, const typename MapType::key_type& key)
+{
+    static_assert(std::is_floating_point_v<typename MapType::key_type>, "Map key type must be floating-point");
+
+    constexpr double tolerance = constants::ERROR_MARGIN;
+
+    auto compare_keys = [&key, &tolerance](const auto& pair) { return std::abs(pair.first - key) < tolerance; };
+
+    return std::find_if(map.cbegin(), map.cend(), compare_keys);
+}
```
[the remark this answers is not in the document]
Drewniok: That's not possible for ``find_key_with_tolerance`` since it is used for ``sidb_energy_distribution``.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: I mean, we already use ``static_assert`` in ``operational_domain_flood_fill``, for example, where we restrict that scenario.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: Sure! We can do this.

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+    /**
+     * Returns the entire operational domain.
+     *
+     * @return The operational domain as a map.
+     */
+    [[nodiscard]] const auto& get_domain() const
+    {
+        return domain_values;
+    }
```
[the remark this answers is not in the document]
Drewniok: Good point and hint! Thanks

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: Good idea! I am not sure if I understand this correctly: I would just add the corresponding function as a private function to the corresponding operational domain, right?

Drewniok: I put it inside ``sidb_simulation_domain`` as a member function.

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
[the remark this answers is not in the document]
Drewniok: I have created ``map_utils.hpp`` where we can put function like ``find_key_with_tolerance``.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -386,36 +317,39 @@ class operational_domain_impl
             canvas_lyt.assign_cell_type(c, technology<Lyt>::cell_type::NORMAL);
         }
 
-        op_domain.dimensions.reserve(num_dimensions);
+        if constexpr (has_dimensions_v<OpDomain>)
```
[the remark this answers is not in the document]
Drewniok: it is not needed anymore.

include/fiction/io/write_operational_domain.hpp
```diff
@@ -145,32 +145,33 @@ inline void write_operational_domain(const OpDomain& opdom, std::ostream& os,
...
+                }
+                else if (num_dimensions == 2)
+                {
+                    writer.write_line(sim_param.parameters[0], sim_param.parameters[1], tag, std::get<1>(op_val));
+                }
+                else  // num_dimensions == 3
+                {
+                    writer.write_line(sim_param.parameters[0], sim_param.parameters[1], sim_param.parameters[2], tag,
+                                      std::get<1>(op_val));
+                }
```
marcelwa, line 173: I know that this code isn't new to this PR, but I just wonder whether we should spell out the first three dimensions here explicitly with an `else` that throws an error. Would make it more robust against future changes and would remind us to fix it in time rather than providing wrong output.

Analogous for the other domain writers.

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Counts the number of key-value pairs in the operational domain.
+     *
+     * @return The number of key-value pairs in the domain.
+     */
+    [[nodiscard]] std::size_t number_of_values() const
+    {
+        return domain_values.size();
+    }
```
marcelwa, line 81: Why not just call it `size()`?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -654,6 +654,7 @@ class operational_domain_impl
      * which the given CDS is physically valid, it is determined whether the CDS is the ground state or the n-th excited
      * state.
      *
+     * @tparam OpDomainTarget The target operational domain type.
```
marcelwa, line 657: Why is this type different from `OpDomain`?

include/fiction/utils/map_utils.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include "fiction/technology/constants.hpp"
+
+#include <algorithm>
+
+namespace fiction
+{
+
+/**
+ * This function searches for a floating-point value specified by the `key` in the provided map `map`, applying a
+ * tolerance specified by `fiction::physical_constants::ERROR_MARGIN`. Each key in the map is compared to the
```
marcelwa, line 17:
```suggestion
 * tolerance specified by `fiction::constants::ERROR_MARGIN`. Each key in the map is compared to the
```

marcelwa (changes requested):
Thank you very much for the restructuring. I think it's already so much cleaner. I spotted some locations where I'd like to offer feedback to hopefully improve upon it even further. Hopefully, this is helpful to you.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -1618,9 +1571,57 @@ class operational_domain_impl
...
+[[nodiscard]] operational_domain operational_domain_grid_search(const Lyt& lyt, const std::vector<TT>& spec,
+                                                                const operational_domain_params& params = {},
+                                                                operational_domain_stats*        stats  = nullptr)
+{
+    static_assert(is_cell_level_layout_v<Lyt>, "Lyt is not a cell-level layout");
+    static_assert(has_sidb_technology_v<Lyt>, "Lyt is not an SiDB layout");
+    static_assert(kitty::is_truth_table<TT>::value, "TT is not a truth table");
+
+    // this may throw an `std::invalid_argument` exception
+    detail::validate_sweep_parameters(params);
```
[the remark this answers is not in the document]
Drewniok: Thank you for this hint!

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,122 @@
+//
+// Created by Jan Drewniok on 29.01.25.
+//
+
+#include "fiction/utils/phmap_utils.hpp"
+
+#include <cstdio>
+#include <functional>
+#include <optional>
+#include <tuple>
```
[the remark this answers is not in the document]
Drewniok: oh, thanks!

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,122 @@
...
+     * Retrieves the value associated with the provided key from the operational domain.
+     * If the key is found, the value is returned; otherwise, `std::nullopt` is returned.
+     *
+     * @param key The key to look up in the domain.
+     * @return The `std::tuple`` associated with the provided key is returned, `std::nullopt` otherwise.
+     */
+    [[nodiscard]] std::optional<std::tuple<MappedTypes...>> get_value(const Key& key) const
+    {
+        return contains_key(key);
+    }
```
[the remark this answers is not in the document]
Drewniok: true

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
Drewniok: Good idea!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: This function returns ``sidb_simulation_domain<parameter_point, uint64_t>``. I think it is unnecessary to make it into a template function. I have changed it.

include/fiction/utils/map_utils.hpp
Drewniok: Good catch!

include/fiction/algorithms/simulation/sidb/sidb_simulation_domain.hpp
```diff
@@ -0,0 +1,122 @@
...
+        return domain_values.size();
+    }
+    /**
+     * Iterates over all key-value pairs in the operational domain and applies a provided callback function
+     * to each pair. This method ensures thread-safe access to the underlying data by leveraging the
+     * `for_each` method of the thread-safe `locked_parallel_flat_hash_map`.
+     *
+     * @param callback A callable object (e.g., lambda, function pointer) that accepts two arguments:
+     *                 the key (`const Key&`) and the associated tuple of mapped values
+     *                 (`const std::tuple<MappedTypes...>&`).
```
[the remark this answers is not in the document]
Drewniok: Thank you!

docs/io/physical_simulation.rst
```diff
@@ -156,6 +156,7 @@ SiDB Operational Domain
         .. autoclass:: mnt.pyfiction.write_operational_domain_params
             :members:
         .. autofunction:: mnt.pyfiction.write_operational_domain
```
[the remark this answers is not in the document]
Drewniok: Thanks!

include/fiction/io/write_operational_domain.hpp
Drewniok: good idea!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
+    /**
+     * Returns the dimensions to sweep over, ordered by priority. The first dimension is the x dimension, the second
+     * dimension is the y dimension, etc.
+     *
+     * @return The dimensions to sweep over.
+     */
+    [[nodiscard]] const std::vector<sweep_parameter>& get_dimensions() const noexcept
+    {
+        return dimensions;
+    }
```
marcelwa, line 208: It's usually also a bit weird to make the underlying container fully accessible like that. We offer a `get` function already for individual access. Additionally, we could provide a subscript operator, i.e., `operator[]` to access dimension values at runtime rather than compile time. In conjunction with `size()`, this can then be used in a loop.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
+    /**
+     * Returns the number of dimensions to sweep over.
+     *
+     * @return The number of dimensions to sweep over.
+     */
+    [[nodiscard]]
+    std::size_t get_number_of_dimensions() const noexcept
+    {
+        return dimensions.size();
+    }
```
marcelwa, line 218: `size()`?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
      */
     std::vector<sweep_parameter> dimensions{};
+};
+/**
+ * The `critical_temperature_domain` class collects the critical temperature for a range of different physical
+ * parameters. It allows for the evaluation of how the critical temperature depends on variations in the underlying
+ * parameter points. This enables simulations to explore the critical temperature's behavior across different conditions
+ * and configurations.
+ */
+class critical_temperature_domain : public sidb_simulation_domain<parameter_point, operational_status, double>
```
marcelwa, line 233: It also tracks the `operational_status`. Maybe that should be mentioned in the docstring also?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
+     * Returns the dimensions to sweep over, ordered by priority. The first dimension is the x dimension, the second
+     * dimension is the y dimension, etc.
+     *
+     * @return The dimensions to sweep over.
+     */
+    [[nodiscard]] const std::vector<sweep_parameter>& get_dimensions() const noexcept
     {
-        operational_values.try_emplace(key, value);
+        return dimensions;
     }
```
marcelwa, line 265: See above

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
+     * Returns the number of dimensions to sweep over.
+     *
+     * @return The number of dimensions to sweep over.
+     */
+    [[nodiscard]]
+    std::size_t get_number_of_dimensions() const noexcept
     {
-        return operational_values;
+        return dimensions.size();
     }
```
marcelwa, line 275: See above

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -721,16 +758,19 @@ class operational_domain_impl
      * @param lyt SiDB cell-level layout that is simulated and compared to the given CDS.
      * @return All physically valid physical parameters and the excited state number.
      */
-    [[nodiscard]] operational_domain<parameter_point, uint64_t>
+    [[nodiscard]] sidb_simulation_domain<parameter_point, uint64_t>
     grid_search_for_physically_valid_parameters(Lyt& lyt) noexcept
```
marcelwa, line 762: Is this the correct return type still with the new domain types available now?

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -1791,10 +1994,70 @@ operational_domain_flood_fill(const Lyt& lyt, const std::vector<TT>& spec, const
...
+ * @param samples Number of samples to perform.
+ * @param params Operational domain computation parameters.
+ * @param stats Operational domain computation statistics.
+ * @return The (partial) operational domain of the layout.
+ */
+template <typename Lyt, typename TT>
+[[nodiscard]] critical_temperature_domain
+critical_temperature_domain_contour_tracing(const Lyt& lyt, const std::vector<TT>& spec, const std::size_t samples,
+                                            const operational_domain_params& params = {},
+                                            operational_domain_stats*        stats  = nullptr)
```
marcelwa, line 2060: This definition is not in the right place yet

include/fiction/io/write_operational_domain.hpp
marcelwa, line 1: Since the `write_*` functions are templates now, their `inline` qualifiers are redundant

include/fiction/io/write_operational_domain.hpp
```diff
@@ -128,21 +129,25 @@ inline void write_operational_domain(const OpDomain& opdom, std::ostream& os,
...
-                              detail::sweep_parameter_to_string(opdom.dimensions[2]), "operational status",
-                              "critical temperature");
+            writer.write_line(
+                detail::sweep_parameter_to_string(dimensions[0]), detail::sweep_parameter_to_string(dimensions[1]),
+                detail::sweep_parameter_to_string(dimensions[2]), "operational status", "critical temperature");
+        }
+
+        else
+        {
+            throw std::invalid_argument("Unsupported number of dimensions: " + std::to_string(num_dimensions));
```
marcelwa, line 150: Use `fmt::format` for string formatting

include/fiction/io/write_operational_domain.hpp
```diff
@@ -157,38 +162,43 @@ inline void write_operational_domain(const OpDomain& opdom, std::ostream& os,
...
-                              detail::sweep_parameter_to_string(opdom.dimensions[1]),
-                              detail::sweep_parameter_to_string(opdom.dimensions[2]), "operational status");
+            writer.write_line(detail::sweep_parameter_to_string(dimensions[0]),
+                              detail::sweep_parameter_to_string(dimensions[1]),
+                              detail::sweep_parameter_to_string(dimensions[2]), "operational status");
+        }
+
+        else
+        {
+            throw std::invalid_argument("Unsupported number of dimensions: " + std::to_string(num_dimensions));
```
marcelwa, line 201: See above

bindings/mnt/pyfiction/include/pyfiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -28,21 +28,49 @@ void operational_domain(pybind11::module& m)
...
+          py::arg("lyt"), py::arg("spec"), py::arg("samples"), py::arg("params") = fiction::operational_domain_params{},
           py::arg("stats") = nullptr, DOC(fiction_operational_domain_flood_fill));
 
-    m.def("operational_domain_contour_tracing", &fiction::operational_domain_contour_tracing<Lyt, py_tt>,
-          py::arg("lyt"), py::arg("spec"), py::arg("samples"), py::arg("params") = fiction::operational_domain_params{},
-          py::arg("stats") = nullptr, DOC(fiction_operational_domain_contour_tracing));
+    m.def(fmt::format("operational_domain_contour_tracing").c_str(),
+          &fiction::operational_domain_contour_tracing<Lyt, py_tt>, py::arg("lyt"), py::arg("spec"), py::arg("samples"),
+          py::arg("params") = fiction::operational_domain_params{}, py::arg("stats") = nullptr,
+          DOC(fiction_operational_domain_contour_tracing));
```
marcelwa, line 47: Lots of redundant `fmt::format` usages

bindings/mnt/pyfiction/include/pyfiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -54,7 +82,9 @@ inline void operational_domain(pybind11::module& m)
     py::class_<fiction::parameter_point>(m, "parameter_point", DOC(fiction_parameter_point))
         .def(py::init<>(), DOC(fiction_parameter_point_parameter_point))
         .def(py::init<const std::vector<double>>(), py::arg("values"), DOC(fiction_parameter_point_parameter_point_2))
-        .def_readwrite("parameters", &fiction::parameter_point::parameters, DOC(fiction_parameter_point))
+
+        .def("set_parameters", &fiction::parameter_point::set_parameters, py::arg("values"))
+        .def("get_parameters", &fiction::parameter_point::get_parameters)
```
marcelwa, line 87: Missing docstrings

bindings/mnt/pyfiction/include/pyfiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -71,28 +101,38 @@ inline void operational_domain(pybind11::module& m)
 
         ;
 
-    py::class_<fiction::operational_domain<fiction::parameter_point, fiction::operational_status>>(
-        m, "operational_domain", DOC(fiction_operational_domain))
+    // todu update docu
```
marcelwa, line 104:
```suggestion
    // todo update docu
```

marcelwa: Please address todos

bindings/mnt/pyfiction/include/pyfiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -71,28 +101,38 @@ inline void operational_domain(pybind11::module& m)
...
+                 self.for_each(
+                     [&py_callback](const auto& key, const auto& value)
+                     {
+                         py_callback(key, value);  // Pass the key and value to the Python callback
+                     });
+             })
+
+        ;
+
+    // todo add docu
```
marcelwa, line 126: Please address todos

marcelwa: CLion should warn you when you try to push code that contains todos

bindings/mnt/pyfiction/include/pyfiction/algorithms/simulation/sidb/physically_valid_parameters.hpp
```diff
@@ -37,21 +37,20 @@ inline void physically_valid_parameters(pybind11::module& m)
 {
     namespace py = pybind11;
 
-    py::class_<fiction::operational_domain<fiction::parameter_point, uint64_t>>(m, "physically_valid_parameters_domain")
+    py::class_<fiction::sidb_simulation_domain<fiction::parameter_point, uint64_t>>(
```
marcelwa, line 40: Missing docstrings for the class and member functions

cli/cmd/simulation/opdom.hpp
```diff
@@ -399,14 +399,14 @@ class opdom_command : public command
...
+    fiction::operational_domain op_domain{};
     /**
      * Writes the operational domain to the specified CSV file.
      */
     void write_op_domain()
     {
-        // if the operational domain call was unsuccessful, do not attempt writing anything
-        if (op_domain.operational_values.empty())
+        // if the operational domain call was unsuccessful, do not attempt to write anything
+        if (op_domain.size() == 0)
```
marcelwa, line 409: Seeing the need for it, it would make sense to provide an `empty` function as well

include/fiction/io/write_defect_influence_domain.hpp
marcelwa, line 1: Same here, function templates don't need to be `inline`

marcelwa (changes requested):
Thanks a lot for addressing my comments so swiftly. The code is a lot cleaner already. Well done! I think, after another iteration, we should be ready to merge :slightly_smiling_face:

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -122,6 +118,30 @@ struct parameter_point
...
     }
+    /**
+     * Sets the parameter values for each dimension.
+     *
+     * @param values The parameter values to set.
+     */
+    void set_parameters(const std::vector<double>& values)
+    {
+        parameters = values;
+    }
```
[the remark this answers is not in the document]
Drewniok: I think I will not provide this function. Instead, the parameters must be set when creating the object.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
+    /**
+     * Sets the dimensions to sweep over, ordered by priority. The first dimension is the x dimension, the second
+     * dimension is the y dimension, etc.
+     *
+     * @param dims The dimensions to sweep over.
+     */
+    void set_dimensions(const std::vector<sweep_parameter>& dims)
+    {
+        dimensions = dims;
+    }
```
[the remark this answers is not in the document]
Drewniok: You are absolutely right! Thanks!

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
```diff
@@ -161,46 +172,116 @@ std::optional<typename MapType::mapped_type> contains_key(const MapType& map, co
...
-        return operational_values;
+        return dimensions.size();
     }
+
+  private:
+    /**
+     * The dimensions to sweep over, ordered by priority. The first dimension is the x dimension, the second dimension
+     * is the y dimension, etc.
+     */
+    std::vector<sweep_parameter> dimensions{};
```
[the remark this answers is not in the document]
Drewniok: not really.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: is already used for `` domain_values.size();`` in ``sidb_simulation_domain``.

include/fiction/algorithms/simulation/sidb/operational_domain.hpp
Drewniok: yes! This is the special case for ``physically_valid_parameters``.

include/fiction/io/write_operational_domain.hpp
Drewniok: Good catch! Thanks!

Drewniok:
> Thanks a lot for addressing my comments so swiftly. The code is a lot cleaner already. Well done! I think, after another iteration, we should be ready to merge 🙂

Thank you so much for your quick and extensive support!

bindings/mnt/pyfiction/include/pyfiction/algorithms/simulation/sidb/physically_valid_parameters.hpp
Drewniok: The problem is that the docstring is not generated. Do you have an idea why?

marcelwa: Unfortunately not. I have observed this behavior a few times myself as well. It's really mysterious to me :thinking:

Drewniok: okay, can we merge it anyway?

Drewniok:
I guess we can merge this one, @marcelwa

marcelwa (approved):

PR #652, opened 2025-01-30
🩹 Non-critical fixes w.r.t. the `ClusterComplete` merge (#390)

wlambooy:
## Description

This PR resolves two non-critical issues that were discovered shortly after merging #390 .

## Checklist:


experiments/equivalence_checking_exact_simulation/equivalence_checking_exact_simulation.cpp
```diff
@@ -94,6 +91,9 @@ int main()  // NOLINT
                     }
 
 #if (FICTION_ALGLIB_ENABLED)
+                    clustercomplete_params cc_params{params};
```
Drewniok, line 94: We should also put ``std::mutex mutex_cc{};`` here.

wlambooy: That does not work. The mutex needs to be defined outside the definition of the threaded function. It's possible to make more `#if` blocks but they are not necessary for successful compilation. I personally think it's alright like this.

Drewniok: sorry, you are right!

experiments/equivalence_checking_exact_simulation/equivalence_checking_exact_simulation.cpp
```diff
@@ -94,6 +91,9 @@
...
```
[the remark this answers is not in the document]
wlambooy: fixed

experiments/equivalence_checking_exact_simulation/equivalence_checking_exact_simulation.cpp
```diff
@@ -94,6 +91,9 @@ int main()  // NOLINT
                     }
 
 #if (FICTION_ALGLIB_ENABLED)
+                    clustercomplete_params<cell<sidb_100_cell_clk_lyt>> cc_params{params};
+                    cc_params.available_threads = 1;
```
Drewniok, line 95: Is there a reason why we only use one thread here?

wlambooy: yes, the parallelisation is performed on the different layouts to simulate.

Drewniok: alright! Thanks!

Drewniok (approved):
Many thanks! 🙏

PR #653, opened 2025-01-30
:bug: Fixed layout resizing in `apply_gate_library`

wlambooy:
## Description

When applying a gate library to a hexagonal gate-level layout, the resulting cell-level layout would not have the right sizing.

The implemented fix now supports tiling geometries in which all even rows/columns are aligned and all odd rows/columns are aligned.

## Checklist:

### DIFF
--- a/include/fiction/algorithms/physical_design/apply_gate_library.hpp
+++ b/include/fiction/algorithms/physical_design/apply_gate_library.hpp
@@ -1,4 +1,4 @@
-asdf//
+//
 // Created by marcel on 28.06.21.
 //
 
--- a/include/fiction/algorithms/physical_design/apply_gate_library.hpp
+++ b/include/fiction/algorithms/physical_design/apply_gate_library.hpp
@@ -40,7 +40,7 @@ class apply_gate_library_impl
   public:
     explicit apply_gate_library_impl(const GateLyt& lyt) :
             gate_lyt{lyt},
-            cell_lyt{get_aspect_ratio_for_cell_level_layout(gate_lyt)}
+            cell_lyt{determine_aspect_ratio_for_cell_level_layout(gate_lyt)}
     {
         cell_lyt.set_tile_size_x(GateLibrary::gate_x_size());
         cell_lyt.set_tile_size_y(GateLibrary::gate_y_size());
--- a/include/fiction/algorithms/physical_design/apply_gate_library.hpp
+++ b/include/fiction/algorithms/physical_design/apply_gate_library.hpp
@@ -221,7 +221,7 @@ class apply_gate_library_impl
         const std::function<cell<CellLyt>(GateLyt, tile<GateLyt>, cell<CellLyt>)> rel_to_abs_cell_pos =
             relative_to_absolute_cell_position<GateLibrary::gate_x_size(), GateLibrary::gate_y_size(), GateLyt,
                                                CellLyt>;
-      
+
         const cell<CellLyt> max_rel_coord = {GateLibrary::gate_x_size() - 1, GateLibrary::gate_y_size() - 1};
 
         const cell<CellLyt> first_odd_tile = {gate_lyt.x() != 0 ? 1 : 0, gate_lyt.y() != 0 ? 1 : 0};
--- a/include/fiction/algorithms/physical_design/apply_gate_library.hpp
+++ b/include/fiction/algorithms/physical_design/apply_gate_library.hpp
@@ -1,4 +1,4 @@
-//
+asdf//
 // Created by marcel on 28.06.21.
 //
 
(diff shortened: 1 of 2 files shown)


include/fiction/algorithms/physical_design/apply_gate_library.hpp
```diff
@@ -207,6 +209,30 @@ class apply_gate_library_impl
...
+     * given gate-level layout, while respecting tiling geometry in which even and odd rows/columns do not line up.
+     *
+     * @param gate_lyt Gate-level layout of which the dimensions are read.
+     * @return Aspect ratio for a cell-level layout that corresponds to the dimensions of the given gate-level layout.
+     */
+    static aspect_ratio<CellLyt> get_aspect_ratio_for_cell_level_layout(const GateLyt& gate_lyt) noexcept
+    {
+        const std::function<cell<CellLyt>(GateLyt, tile<GateLyt>, cell<CellLyt>)> rel_to_abs_cell_pos =
+            relative_to_absolute_cell_position<GateLibrary::gate_x_size(), GateLibrary::gate_y_size(), GateLyt,
+                                               CellLyt>;
```
marcelwa, line 223: It might be possible to write this without an assignment to a function object (untested though)
```suggestion
        using rel_to_abs_cell_pos = relative_to_absolute_cell_position<GateLibrary::gate_x_size(), GateLibrary::gate_y_size(), GateLyt, CellLyt>;
```

include/fiction/algorithms/physical_design/apply_gate_library.hpp
```diff
@@ -207,6 +209,30 @@ class apply_gate_library_impl
...
         }
     }
+    /**
+     * Computes the (inclusively) bounding coordinate for a cell-level layout that is derived from the dimensions of the
+     * given gate-level layout, while respecting tiling geometry in which even and odd rows/columns do not line up.
+     *
+     * @param gate_lyt Gate-level layout of which the dimensions are read.
+     * @return Aspect ratio for a cell-level layout that corresponds to the dimensions of the given gate-level layout.
+     */
+    static aspect_ratio<CellLyt> get_aspect_ratio_for_cell_level_layout(const GateLyt& gate_lyt) noexcept
```
marcelwa, line 219: I'd suggest the name `calculate_aspect_ratio_for_cell_level_layout` (`calculate` instead of `get`) because some actual work is performed. A getter would usually only fetch with no real overhead.

include/fiction/algorithms/physical_design/apply_gate_library.hpp
```diff
@@ -207,6 +209,30 @@ class apply_gate_library_impl
...
+        const cell<CellLyt> max_rel_coord = {GateLibrary::gate_x_size() - 1, GateLibrary::gate_y_size() - 1};
+
+        const cell<CellLyt> first_odd_tile = {gate_lyt.x() != 0 ? 1 : 0, gate_lyt.y() != 0 ? 1 : 0};
+
+        const auto max_coord_even_x = rel_to_abs_cell_pos(gate_lyt, {0, gate_lyt.y()}, max_rel_coord);
+        const auto max_coord_odd_x  = rel_to_abs_cell_pos(gate_lyt, {first_odd_tile.x, gate_lyt.y()}, max_rel_coord);
+        const auto max_coord_even_y = rel_to_abs_cell_pos(gate_lyt, {gate_lyt.x(), 0}, max_rel_coord);
+        const auto max_coord_odd_y  = rel_to_abs_cell_pos(gate_lyt, {gate_lyt.x(), first_odd_tile.y}, max_rel_coord);
+
+        return {std::max(max_coord_even_y.x, max_coord_odd_y.x), std::max(max_coord_even_x.y, max_coord_odd_x.y)};
```
marcelwa, line 234: Since the old functionality worked properly for `cartesian_layout`s, I was wondering if you considered working with an `if constexpr` condition here to distinguish between topologies at compile-time.

marcelwa (changes requested):
Many thanks for the quick fix! Some minor comments below :slightly_smiling_face:

include/fiction/algorithms/physical_design/apply_gate_library.hpp
wlambooy: I went for `determine` 👍

include/fiction/algorithms/physical_design/apply_gate_library.hpp
wlambooy: It's not the most efficient to do these four function calls, but I'm quite sure that they are so inexpensive that it won't matter in practise. Let me know if you think differently.

marcelwa: That separation of concerns makes sense to me. And I agree that it should be fine as it is also not a time-critical function. Plus, it is only executed once per layout.

wlambooy:
It looks like Clang-Tidy is not too happy about `using` with something that is not a type

marcelwa (approved):

PR #685, opened 2025-03-03
:bug: Increased tolerance to avoid undetected degeneracy and added forbidden BDL types in wire detection

Drewniok:
## Description

This PR increases the tolerance used in ``groundstate_from_simulation_results`` to extract the ground states from simulation results. Previously, it failed to detect the degeneracy of the ground state in some cases, leading to "weird" operational domains. This should now be fixed. 

Example:
This operational domain should be empty, which is only the case if the degeneracy of the ground state is correctly detected. As said, this should be fixed now.
perturber_distance_encoded_two_bdl_wire

## Checklist:


marcelwa:
Thanks for the fix! :pray:

Slightly off-topic, but thinking about the function, I wonder if we should refactor it into a member of `sidb_simulation_result` called `groundstate` for more intuitive usability. What do you think, @Drewniok and @wlambooy?

Drewniok:
> Thanks for the fix! 🙏
> 
> Slightly off-topic, but thinking about the function, I wonder if we should refactor it into a member of `sidb_simulation_result` called `groundstate` for more intuitive usability. What do you think, @Drewniok and @wlambooy?

That's a great idea! I like it!

wlambooy:
> Thanks for the fix! :pray:
> 
> Slightly off-topic, but thinking about the function, I wonder if we should refactor it into a member of `sidb_simulation_result` called `groundstate` for more intuitive usability. What do you think, @Drewniok and @wlambooy?

I think that is an excellent idea! Since there is always the possibility of getting degenerate groundstates, we should respect that in the function name, i.e.: `groundstates`.

marcelwa:
> > Thanks for the fix! :pray:
> > 
> > Slightly off-topic, but thinking about the function, I wonder if we should refactor it into a member of `sidb_simulation_result` called `groundstate` for more intuitive usability. What do you think, @Drewniok and @wlambooy?
> 
> I think that is an excellent idea! Since there is always the possibility of getting degenerate groundstates, we should respect that in the function name, i.e.: `groundstates`.

Good call! Would it make sense then to parameterize the function to allow degeneracy or not? Or have two separate functions for that? I'm also okay with having a single one simply called `groundstates`. I'm merely asking for your preference since you're going to use the API the most.

include/fiction/algorithms/simulation/sidb/detect_bdl_wires.hpp
```diff
@@ -597,11 +597,13 @@ class detect_bdl_wires_impl
...
      *
      * @param type The type of the cell to filter by.
+     * @param forbidden_type The type of cell that is forbidden in the wire and is automatically filtered out.
      * @return A vector of `bdl_wire` objects containing cells of the specified type. If no such wires are found,
      *         an empty vector is returned.
      */
     [[nodiscard]] std::vector<bdl_wire<Lyt>>
-    filter_wires_by_type(const typename technology<Lyt>::cell_type& type) const noexcept
+    filter_wires_by_type(const typename technology<Lyt>::cell_type& type,
+                         const typename technology<Lyt>::cell_type  forbidden_type) const noexcept
```
marcelwa, line 606: Maybe I'm missing something here. If you filter wires by a certain type, why do you need to specify another type that is prohibited? Wouldn't any type `!= type` be automatically prohibited by definition?

marcelwa, line 606: What if I want to prohibit multiple types?

marcelwa, line 606: Why is one type a reference and the other a copy?

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
```diff
@@ -59,6 +63,59 @@ struct sidb_simulation_result
...
+
+        // Find the minimum energy
+        double min_energy = std::numeric_limits<double>::infinity();
+        if (!charge_configurations_copy.empty())
+        {
+            min_energy = std::min_element(charge_configurations_copy.begin(), charge_configurations_copy.end(),
+                                          [](const auto& lhs, const auto& rhs)
+                                          { return lhs.get_system_energy() < rhs.get_system_energy(); })
+                             ->get_system_energy();
+        }
```
marcelwa, line 99: Don't we have an existing function for minimum energy?

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
```diff
@@ -59,6 +63,59 @@ struct sidb_simulation_result
...
+
+        for (const auto charge_index : charge_indices)
+        {
+            const auto cds_it =
+                std::find_if(charge_configurations_copy.cbegin(), charge_configurations_copy.cend(),
+                             [&](const auto& cds)
+                             {
+                                 return cds.get_charge_index_and_base().first == charge_index &&
+                                        std::abs(cds.get_system_energy() - min_energy) < constants::ERROR_MARGIN;
+                             });
```
marcelwa, line 109: Could you add a comment like above?

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
```diff
@@ -59,6 +63,59 @@ struct sidb_simulation_result
...
+
+        // in case, the charge indices are not updated.
+        auto charge_configurations_copy = charge_distributions;
+
+        // Find all unique charge indices. This is done because simulation results can have multiple identical charge
+        // distributions.
+        for (auto& cds : charge_configurations_copy)
+        {
+            cds.charge_distribution_to_index();
+            charge_indices.insert(cds.get_charge_index_and_base().first);
```
marcelwa, line 88: If we're relying on the charge indices here, wouldn't that break for > 64 dots?

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
```diff
@@ -59,6 +63,59 @@ struct sidb_simulation_result
      * The key of the map is the name of the parameter, the element is the value of the parameter.
      */
     std::unordered_map<std::string, std::any> additional_simulation_parameters{};
+    /**
+     * This function calculates the ground state charge distributions from the provided simulation results.
+     * The ground state charge distributions are those with energy closest to the minimum energy found in the simulation
+     * results.
```
marcelwa, line 69:
```suggestion
    /**
     * This function calculates the ground state charge distributions from the stored simulation results.
     * The ground state charge distributions are those with energy closest to the minimum energy found in the simulation
     * results.
```

marcelwa, line 69: What do we mean by "energy closest to the minimum found in the simulation results"? Don't we return **the** minimum ones?

test/algorithms/simulation/sidb/quickexact.cpp
```diff
@@ -2043,7 +2042,7 @@ TEMPLATE_TEST_CASE("QuickExact gate simulation of Si-111 surface", "[quickexact]
 
     const auto simulation_results = quickexact<TestType>(lyt, params);
 
-    const auto ground_state = groundstate_from_simulation_result(simulation_results);
+    const auto ground_state = simulation_results.get_groundstates();
```
marcelwa, line 2045: Is it not called `groundstates()`? How does this compile?

test/algorithms/simulation/sidb/quickexact.cpp
```diff
@@ -2067,7 +2066,7 @@ TEMPLATE_TEST_CASE("QuickExact AND gate simulation of Si-111 surface", "[quickex
 
         const auto simulation_results = quickexact<TestType>(lyt, params);
 
-        const auto ground_state = groundstate_from_simulation_result(simulation_results);
+        const auto ground_state = simulation_results.get_groundstate();
```
marcelwa, line 2069: See above

test/algorithms/simulation/sidb/quickexact.cpp
```diff
@@ -2108,7 +2107,7 @@ TEMPLATE_TEST_CASE("QuickExact AND gate simulation of Si-111 surface", "[quickex
         const auto simulation_results = quickexact<TestType>(lyt, params);
         CHECK(simulation_results.charge_distributions.size() == 7);
 
-        const auto ground_state = groundstate_from_simulation_result(simulation_results);
+        const auto ground_state = simulation_results.get_groundstates();
```
marcelwa, line 2110: See above

include/fiction/algorithms/simulation/sidb/detect_bdl_wires.hpp
Drewniok: Yes, I understand that this is not great and confusing. Thanks for bringing this up!
I think we should solve this differently and use a different name for the input. The idea is that ``forbidden_type`` can be used to remove BDL pairs of the given type from the wire. This is needed in the specific case of a BDL wire with input and output BDL pairs (typically the BDL wires of our gates have either input or output BDL wires and not both).

Drewniok: To be honest, I don't see a use for that at the moment.

Drewniok: Good catch!

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
Drewniok: You are absolutely right! This is still old code. Thanks!

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
Drewniok: That is a really good point and something we should be aware of. And we had this discussion over a year ago that we need to replace the ``uint64_t`` charge index with something else in the future. 
So far I have been doing simulations for smaller layouts. So we could live with it. However, as @wlambooy takes big steps towards larger layouts, this may become an issue at some point. Typically our SiDB circuit has about 50% negatively charged SiDBs, which means it could break at about 120 dots, right?

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
Drewniok: good point! It is misleading here. I will change it! Thanks.

include/fiction/algorithms/simulation/sidb/detect_bdl_wires.hpp
Drewniok: hmm, what do you need about the new version.

marcelwa (changes requested):

include/fiction/algorithms/simulation/sidb/detect_bdl_wires.hpp
marcelwa: I like it better. Thank you!

include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
```diff
@@ -59,6 +64,55 @@ struct sidb_simulation_result
      * The key of the map is the name of the parameter, the element is the value of the parameter.
      */
     std::unordered_map<std::string, std::any> additional_simulation_parameters{};
+    /**
+     * This function computes the ground state of the charge distributions.
+     *
+     * @note When degenerate states exist, there are multiple ground states with the same energy.
```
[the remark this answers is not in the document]
Drewniok: Thank you!

include/fiction/algorithms/simulation/sidb/detect_bdl_wires.hpp
```diff
@@ -606,24 +606,70 @@ class detect_bdl_wires_impl
         std::vector<bdl_wire<Lyt>> filtered_wires{};
         std::optional<std::size_t> wire_length_of_the_first_wire{};  // Track the length of the first wire
 
+        // Filter out the BDL pairs of the opposite type. This is needed if a wire contains both input and output cells.
+        // Example: If an input wire is selected, the output cells are removed and vice versa.
+        std::optional<typename technology<Lyt>::cell_type> filtered_out_bdl_pair{};
```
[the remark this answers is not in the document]
Drewniok: True! Thank you.

marcelwa (approved):
Many thanks for the fix and the quick iterations!

PR #721, opened 2025-04-24
:bug: Use correct function for charge index calculation and fix base for TTS calculation

Drewniok:
## Description

This PR fixes two bugs: 

1.) The charge index calculation was done wrongly for ``groundstates()`` for charge distributions with base = 3. This is now fixed by using ``charge_distribution_to_index_general()``. 
2.) The automatic base detection in the TTS calculation was activated. Hence, the heuristic and exact simulations might be conducted with different bases. This is not desired. The automatic base detection is now turned off, ensuring that the same base is used in both cases. 

## Checklist:

### DIFF
--- a/include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
+++ b/include/fiction/algorithms/simulation/sidb/sidb_simulation_result.hpp
@@ -81,7 +81,7 @@ struct sidb_simulation_result
         // distributions.
         for (auto& cds : charge_distributions)
         {
-            cds.charge_distribution_to_index();
+            cds.charge_distribution_to_index_general();
             charge_indices.insert(cds.get_charge_index_and_base().first);
         }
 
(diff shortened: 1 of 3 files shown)


marcelwa (approved):
Thanks for the fix! Could you indicate in the PR description, what the previous behavior entailed? For the base detection, it is also not clear to me if you're talking about the old or the fixed behavior. There seems to be a bit of an ambiguity.

Drewniok:
> Thanks for the fix! Could you indicate in the PR description, what the previous behavior entailed? For the base detection, it is also not clear to me if you're talking about the old or the fixed behavior. There seems to be a bit of an ambiguity.

sure! is this better now?

marcelwa:
> > Thanks for the fix! Could you indicate in the PR description, what the previous behavior entailed? For the base detection, it is also not clear to me if you're talking about the old or the fixed behavior. There seems to be a bit of an ambiguity.
> 
> sure! is this better now?

Yes, it's great. Thanks!

PR #730, opened 2025-04-29
:bug: Fixed not setting Z dimension while determining aspect ratio for cell-level layout

wlambooy:
## Description

Thanks a lot to @simon1hofmann for reporting this bug! For SiDB layouts, there is no issue, but it can be observed when, for instance, the QCAOne library is applied to a gate-level layout with crossings. In PR #653 , some code was changed and it was overlooked to think about setting the Z dimension.

## Checklist:


simon1hofmann:
Thanks a lot for the quick fix 🙏

marcelwa (changes requested):
Thank you for the fix, @wlambooy! :pray: When we notice a bug and want to address it, we usually start with writing a reproducing test case. This approach will, first, ensure the bug is gone and, second, prevent it from re-occurring. Would you thus please add a respective unit test?

marcelwa:
@wlambooy could you address `clang-tidy`'s suggestions, please?

test/algorithms/physical_design/apply_gate_library.cpp
```diff
@@ -562,3 +565,30 @@ TEST_CASE("Gate-level layout with with different gates", "[apply-gate-library]")
         }
     }
 }
+
+TEST_CASE("Applying the QCAOne gate library", "[apply-gate-library]")
```
marcelwa, line 569:
```suggestion
TEST_CASE("Applying the QCA ONE gate library", "[apply-gate-library]")
```

test/algorithms/physical_design/apply_gate_library.cpp
```diff
@@ -562,3 +565,30 @@ TEST_CASE("Gate-level layout with with different gates", "[apply-gate-library]")
...
+
+    SECTION("Without crossings")
+    {
+        const auto gate_lyt = blueprints::straight_wire_gate_layout<GateLyt>();
+
+        const auto layout = apply_gate_library<qca_cell_clk_lyt, qca_one_library, GateLyt>(gate_lyt);
+
+        CHECK(layout.x() > 0);
+        CHECK(layout.y() > 0);
+        CHECK(layout.z() == 0);
```
marcelwa, line 581: These checks are not very specific. They can be made a lot stricter.

test/algorithms/physical_design/apply_gate_library.cpp
```diff
@@ -562,3 +565,30 @@ TEST_CASE("Gate-level layout with with different gates", "[apply-gate-library]")
...
+
+    SECTION("With crossings")
+    {
+        const auto gate_lyt = blueprints::optimization_layout_corner_case_outputs_2<GateLyt>();
+
+        const auto layout = apply_gate_library<stacked_qca_cell_clk_lyt, qca_one_library, GateLyt>(gate_lyt);
+
+        CHECK(layout.x() > 0);
+        CHECK(layout.y() > 0);
+        CHECK(layout.z() > 0);
```
marcelwa, line 592: These checks are not very specific. They can be made a lot stricter.

marcelwa (changes requested):

test/algorithms/physical_design/apply_gate_library.cpp
marcelwa: I would argue not. But they can be explicitly specified. Seems more robust, no?

marcelwa (approved):

PR #552, opened 2024-10-30
:sparkles: Sorting designed SiDB gates

wlambooy:
## Description

This PR introduces an extension to `is_operational` and `design_sidb_gates`. The former is now able to return the simulation results that certify the status `OPERATIONAL`, which may then be used by the extension to the latter, which uses the simulation results to return the designed layouts in a specific ordering. This PR implements a simple heuristic for this ordering, namely it prefers the layouts for the the ground state isolation is large, ie., the energetic gap between the ground state and the first excited state. Specifically, it sorts by the minimum ground state isolation for each input.

To use this new feature when designing SiDB gates, make sure that all combinations are enumerated, and set the `post_design_process` parameter to `PREFER_ENERGETICALLY_ISOLATED_GROUND_STATES`.

## Checklist:

- [x] The pull request only contains commits that are related to it.
- [x] I have added appropriate tests and documentation.
- [ ] I have added a changelog entry.
- [x] I have created/adjusted the Python bindings for any new or updated functionality.
- [ ] I have made sure that all CI jobs on GitHub pass.
- [x] The pull request introduces no new warnings and follows the project's style guidelines.

## Summary by CodeRabbit

* **New Features**
  * Added richer SiDB “operational assessment” outputs (overall status plus optional per-input results), including new Python assessment types.
  * Enhanced SiDB gate design configuration with lattice-specific parameter classes (100/111), richer mode/termination controls, and post-design ordering using ground-state isolation.
  * Introduced separate configuration for kink handling vs positive-charge feasibility.
* **Bug Fixes**
  * Improved handling of edge cases where no valid configuration exists (including zero-entity combination generation).
* **Documentation**
  * Updated SiDB gate design, operational-domain, and Python API documentation to match the new outputs and configuration.
* **Tests**
  * Refreshed unit tests for the updated APIs and parameter fields.

PR #747, opened 2025-05-22
:bug: Fixed I/O cell handling in ``assign_cell_type``

Drewniok:
## Description

This PR addresses a bug in the ``assign_cell_type`` function. Previously, when assigning a new cell type to an input or output cell, the corresponding updates to the inputs and outputs were not correctly applied. This issue has now been resolved, ensuring that inputs and outputs are properly updated when a cell type is reassigned.

## Checklist:


marcelwa:
@Drewniok could you make sure that the tests hit 95% of added lines? Then, I will have a look.

include/fiction/algorithms/physical_design/design_sidb_gates.hpp
```diff
@@ -699,6 +703,13 @@ class design_sidb_gates_impl
 
             if (lyt_copy.get_cell_type(all_sidbs_in_canvas[i]) == sidb_technology::cell_type::EMPTY)
             {
+                if constexpr (is_sidb_defect_surface_v<Lyt>)
```
wlambooy, line 706: A while ago for #552 I was already struggling with this function and the one defined below. Can they not be merged into one function? The current way seems very error prone. Also, above, there is `Lyt lyt_copy{skeleton_layout.clone()};` This might perform two deep copies.

include/fiction/algorithms/physical_design/design_sidb_gates.hpp
```diff
@@ -720,15 +731,18 @@ class design_sidb_gates_impl
...
+                // SiDBs cannot be placed on positions which are already occupied by atomic defects.
+                if constexpr (is_sidb_defect_surface_v<Lyt>)
                 {
-                    return std::nullopt;
+                    if (skeleton_layout.get_sidb_defect(all_sidbs_in_canvas[i]).type != sidb_defect_type::NONE)
+                    {
+                        return std::nullopt;
+                    }
                 }
+                lyt.assign_cell_type(all_sidbs_in_canvas[i], sidb_technology::cell_type::LOGIC);
```
wlambooy, line 744: Just an idea: could `sidb_defect_surface` overload `assign_cell_type` and perform the check whether there is already a defect present at that location, and throws a exception if that is the case? Then I think we don't have to work with `std::optional` anymore, which complicates things quite a bit. Then, in some cases, the exception may be caught, for example when we want to assign a canvas SiDB which we allow can be omitted in that case, while we don't catch it for skeleton SiDBs, since we want a nice error message to roll out in that case.

include/fiction/algorithms/physical_design/design_sidb_gates.hpp
```diff
@@ -637,7 +637,11 @@ class design_sidb_gates_impl
 
                     for (std::size_t j = start_index; j < end_index; ++j)
                     {
-                        conduct_pruning_steps(all_canvas_layouts[j]);
+                        const auto canvas_lyt = all_canvas_layouts[j];
+                        if (!canvas_lyt.is_empty())
```
wlambooy, line 641: I think this check should be made in `conduct_pruning_steps`. This connects to my comment on line 744, since I think instead of returning early when `!canvas_lyt.has_value()`, it should return early upon `canvas_lyt.is_empty()`.

include/fiction/layouts/cell_level_layout.hpp
```diff
@@ -167,17 +167,41 @@ class cell_level_layout : public ClockedLayout
             strg->cell_mode_map.erase(c);
             strg->inputs.erase(c);
             strg->outputs.erase(c);
-
             return;
         }
-        if (Technology::is_input_cell(ct))
+
+        const auto is_input  = Technology::is_input_cell(ct);
```
wlambooy, line 173: Currently, I fail to see why the changes here are necessary. Maybe I'm missing something; does this actually change anything? Since each cell can only have one cell type, then together with the assumption that each technology distinguishes between a normal and a input and a output cell type, the original `if .. else if .. else` code should do exactly the same. I think also the `is_pi` or `is_po` checks are not needed since `erase` will just not do anyting if the cell is not found in the respective store, which is exactly what `is_pi`/`is_po` looks for. If you could provide an example where the updated code gives a different result than the original code, that would be great. I ran the extended test included in this PR below without these changes, and all the tests passed.

wlambooy:
Hi Jan, thanks a lot for the continued development here! I added some quite detailed comments of concerns that arose while looking into it, I'm looking forward to hear if you agree! Let's strive for simplicity and longevity in _fiction_! 💪

include/fiction/algorithms/physical_design/design_sidb_gates.hpp
Drewniok: My concern is that this would cause quite some overload, right?

include/fiction/algorithms/physical_design/design_sidb_gates.hpp
Drewniok: sure, I can do it. Thanks.

include/fiction/algorithms/physical_design/design_sidb_gates.hpp
```diff
@@ -720,15 +731,18 @@ class design_sidb_gates_impl
         {
             assert(i < all_sidbs_in_canvas.size() && "cell indices are out-of-range");
 
-            // SiDBs cannot be placed on positions which are already occupied by atomic defects.
-            if constexpr (is_sidb_defect_surface_v<Lyt>)
+            if (skeleton_layout.get_cell_type(all_sidbs_in_canvas[i]) == sidb_technology::cell_type::EMPTY)
```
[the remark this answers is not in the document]
Drewniok: The concern raised in point 3 is entirely valid. However, as it represents a more specific edge case, I believe it would be best to address it in a separate PR. This approach allows us to prioritize merging this urgent bug fix promptly without unnecessary delays. If you like, @wlambooy, please feel free to add it to Nexus as a todo for the future.

include/fiction/layouts/cell_level_layout.hpp
Drewniok: Thank you, @wlambooy! 🙏

Drewniok:
I would be happy to hear your feedback on this PR, @marcelwa. Many thanks! 🙏

marcelwa (approved):
Thank you @Drewniok for the fix and @wlambooy for the initial review. In my opinion, you did a great job, as I don't spot anything to comment on.

PR #1032, opened 2026-08-02
:recycle: Migrate C++ codebase from C++17 to C++20

marcelwa:
## Description

First incremental step in migrating `fiction` from C++17 to C++20: bumps the language standard and applies safe, mechanical C++20 idiom modernizations across the codebase. **Concepts, Modules, and Coroutines are explicitly out of scope** for this pass and are left untouched, to be tackled separately.

### What changed

- **Language standard**: default `CMAKE_CXX_STANDARD` bumped to 20 (root `CMakeLists.txt`, vendored `graph-coloring` lib); docs (`AGENTS.md`, `README.md`) updated to say C++20.
- **Salvaged the old `c++20` spike branch**: cherry-picked its still-relevant, non-Concepts commits (`.clang-format` C++20 style, spaceship-operator and aggregate-init examples) rather than discarding 20 months of prior intent; its Concepts-based commits and one badly stale conflict were dropped.
- **New reference doc**: `docs/cpp20-modernization-guide.md` — a token-efficient cheat sheet (name + example + cppreference link per feature) used to keep the modernization pass consistent and scoped.
- **Modernization pass** across `include/fiction/**` and matching `test/**`, applied per-subdirectory:
  - `std::ranges` algorithms replacing `algo(c.begin(), c.end(), ...)` full-container calls (the largest share of changes)
  - `std::erase`/`std::erase_if` replacing the erase-remove idiom
  - Designated initializers for params-struct construction
  - `operator<=>` / `= default` for comparison-operator pairs, applied only where memberwise comparison is semantically equivalent to the existing hand-written operators (several candidates were deliberately left hand-written where semantics differed)
  - `using enum` to de-verbosify heavily-qualified enum-class usage
- **Build-verification fixes** (found by doing one full build + test run at the end, per plan, rather than compiling after every edit):
  - 3 iterator-type mismatches (`std::ranges::find`/`lower_bound` returning a non-`const` iterator paired with a `.cbegin()` in the following `std::distance` call)
  - 1 `using enum` on an enum declared locally inside a function template (GCC rejects this as a dependent type) — reverted to explicit qualification
  - 2 **pre-existing, unrelated** `fmt` v12 consteval format-string violations that were blocking the build — fixed with `fmt::runtime(...)` since they were already broken on `main`

### Verification

- Full build (`tests-full` preset, Z3 + Mugen + ALGLIB enabled, experiments excluded — see note below): 359/359 targets, zero errors.
- `ctest`: 119/120 passed. The one failure (`one_pass_synthesis`) is caused by a missing Python `graphviz` package required by the optional Mugen SAT-based solver in this environment — confirmed unrelated to this PR (same dependency warning appears at CMake configure time, and the test never reaches C++ logic before failing on the missing module).
- `experiments/` was excluded from the verification build: it currently fails to compile on `main` too, due to the same pre-existing `fmt` v12 consteval issue in `design_rule_violations.hpp` (fixed here) hitting a code path only `experiments/` exercises via a different call site. Worth a follow-up but out of scope for this PR.

## Checklist:

- [x] I have added appropriate tests and documentation. *(no new tests needed — behavior-preserving idiom pass; added the modernization guide as documentation)*
- [ ] I have created/adjusted the Python bindings for any new or updated functionality. *(no functional changes, so N/A)*
- [ ] I have made sure that all CI jobs on GitHub pass. *(not yet run on this environment's CI — please check on push)*
- [x] The pull request introduces no new warnings and follows the project's style guidelines. *(ran through the repo's `clang-format`/pre-commit hooks throughout)*

## Summary by CodeRabbit

* **New Features**
  * The project now requires and supports C++20.
  * Improved cancellation handling for selected long-running design and simulation operations.

* **Documentation**
  * Added a C++20 modernization guide.
  * Updated setup, project overview, and development guidance to reflect the C++20 requirement.

* **Bug Fixes**
  * Improved formatting reliability for runtime-generated diagnostic and layout output.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -82,6 +82,6 @@ SpacesInCStyleCastParentheses: false
 SpacesInContainerLiterals: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
-Standard: c++17
+Standard: c++20
 UseTab: Never
 LineEnding: LF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -36,6 +36,7 @@ Checks: |
   -readability-identifier-length,
   -readability-isolate-declaration,
   -readability-magic-numbers,
+  -readability-math-missing-parentheses,
   -readability-named-parameter,
   -readability-uppercase-literal-suffix
   -readability-redundant-member-init
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,12 +1,12 @@
 # AGENTS.md
 
-You are an expert software architect and engineer specializing in **C++17**, **Python**, and **Field-coupled Nanocomputing (FCN)** design automation. You are working on the `fiction` project.
+You are an expert software architect and engineer specializing in **C++20**, **Python**, and **Field-coupled Nanocomputing (FCN)** design automation. You are working on the `fiction` project.
 
 ## Persona
 
 - **Role**: Core developer and maintainer.
 - **Expertise**:
-  - Modern C++ (C++17 standard).
+  - Modern C++ (C++20 standard).
   - Python bindings using `pybind11`.
   - CMake build systems.
   - FCN technologies (QCA, iNML, SiDB).
@@ -23,7 +23,7 @@ You are an expert software architect and engineer specializing in **C++17**, **P
 ## Project Knowledge
 
 - **Tech Stack**:
-  - **C++**: C++17 (Strict), `clang-format`, `clang-tidy`.
+  - **C++**: C++20 (Strict), `clang-format`, `clang-tidy`.
   - **Python**: Python 3.10+, `pybind11`, `scikit-build-core`, `nox`, `pytest`.
   - **Build System**: CMake 3.23+.
   - **Documentation**: Doxygen.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,8 +6,8 @@ cmake_minimum_required(VERSION 3.23...4.2)
 
 # Only set the CMAKE_CXX_STANDARD if it is not set by someone else
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-  # Set C++ standard; at least C++17 is required
-  set(CMAKE_CXX_STANDARD 17)
+  # Set C++ standard; at least C++20 is required
+  set(CMAKE_CXX_STANDARD 20)
 endif()
 
 # strongly encouraged to enable this globally to avoid conflicts between
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
   </picture>
 </p>
 
-This code base provides a C++17 framework for **fi**eld-**c**oupled **t**echnology-**i**ndependent **o**pen
+This code base provides a C++20 framework for **fi**eld-**c**oupled **t**echnology-**i**ndependent **o**pen
 **n**anocomputing developed as part of the _Munich Nanotech Toolkit_ (_MNT_) by the
 [Chair for Design Automation](https://www.cda.cit.tum.de/) at the [Technical University of Munich](https://www.tum.de/).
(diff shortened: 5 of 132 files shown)


marcelwa:
Pushed a follow-up round addressing three specific asks:

**1. `std::jthread` + `std::stop_token`** — replaced all `std::thread` usage across `include/fiction/**` and `experiments/**`. Where a shared `std::atomic<bool>` flag was being polled by worker threads to signal early termination (3 functions in `design_sidb_gates.hpp`), it's now a proper `std::stop_source`/`stop_token` pair.

**2. `std::ranges` completion** — audited `experiments/**` (nothing applicable — no full-container algorithm calls there) and spot-checked `include/`/`test/` for anything the first pass missed; found and converted a few (an inconsistent reverse-scan in `sidb_cluster_hierarchy.hpp`, a shuffle pattern in `graph_oriented_layout_design.hpp`).

**3. Compound (designated) initialization completion** — extended the same "construct + set N fields" → designated-initializer pattern into `experiments/**` (previously out of scope) and a few more `include/`/`test/` spots.

**Docs** — added a `std::jthread`/`std::stop_token` entry to `docs/cpp20-modernization-guide.md`; updated `docs/about.rst` and `docs/getting_started.rst`, which still described the project as C++17.

**Verification** — this round also enabled `FICTION_EXPERIMENTS=ON` in the build for the first time (previously excluded). That surfaced 3 categories of real issues, all fixed:
- `mockturtle::cut_rewriting_params` and `fiction::sidb_simulation_parameters`/`sidb_simulation_result` are non-aggregate types (user-declared constructors), so designated initializers don't apply to them — reverted those specific sites to construct-then-assign or positional-constructor form.
- One more pre-existing (unrelated to this PR) `fmt` v12 consteval format-string violation in `cell_ports.hpp`, only reachable via an experiments/-only code path.

Full build with experiments enabled: 519/519 targets, zero errors. `ctest`: 119/120 (same pre-existing, unrelated `one_pass_synthesis` failure as before, due to a missing Python `graphviz` package).

marcelwa:
## CodeRabbit review — fixes applied

Addressed all 7 open threads:

1. **`coordinates.hpp` — `cube::coord_t::operator<=>` ordering regression (🟠 Major)**: the defaulted `<=>` compared `d` (dead flag) first, so a live and dead coordinate at the same position stopped being order-equivalent, unlike the original hand-written `operator<`. Hand-wrote `operator<=>` (returning `std::weak_ordering` over z/y/x only, dead flag ignored) plus a separate `operator==` respecting the dead flag — restoring the exact original semantics. Added a regression test.
2. **`stl_utils.hpp` — missing `<algorithm>` include**: added; `std::ranges::find` was only reachable via a transitive include.
3. **`write_location_and_ground_state.cpp` / `write_svg_layout.cpp` — UB passing signed `char` to `::isspace`**: cast to `unsigned char` in both erase_if predicates.
4. **`vendors/graph-coloring/CMakeLists.txt` — modified vendored file (🟠 Major, policy violation)**: reverted to `C++17`; this vendored subproject compiles fine as its own translation units regardless of the parent project's standard, so there was no need to touch it.
5. **Modernization guide — `m.find(k)` example**: noted the `m.end()` precondition.
6. **Modernization guide — `std::jthread` "fire-and-forget" wording**: corrected; the destructor requests stop and joins, it doesn't detach.
7. **Skipped**: the missing-Doxygen finding on `bestagon.cpp`'s `main`, `generate_edge_intersection_graph_impl::run`, and `sat_coloring_handler`'s constructor. Checked — this matches the codebase's established, consistent convention: `experiments/**`'s `main()` functions and internal `*_impl`/handler classes' constructors and `run()` methods are never Doxygen-documented anywhere in the repo (only the public API surface is); these three predate this PR and weren't newly introduced undocumented.

All changes verified by building and running the affected test binaries locally.

PR #1037, opened 2026-08-03
:wastebasket: Deprecate functions from `v0.7.1`

marcelwa:
## Summary

Reintroduces deprecation markers that were originally prepared on `deprecated-since-0.7.0` but never merged before v0.7.0 shipped. Since v0.7.0 is already public without these warnings, the deprecation window is shifted by one patch release: **deprecated since 0.7.1, removed in 0.8.0**.

Nothing is functionally removed here — this only adds `[[deprecated]]` attributes, doc comments, and CLI-help notices to:

- `qca_energy_dissipation()`
- `one_pass_synthesis()`
- `jump_point_search()`
- `range_t`
- `exhaustive_ground_state_simulation()` (C++ and pyfiction binding)

Cherry-picked from the stale `deprecated-since-0.7.0` branch and rebased onto current `main`, resolving conflicts from the CLI build restructuring (#854) that moved several `cli/cmd/**/*.hpp` files into `include/` and `src/` subdirectories.

## Test plan

- [ ] CI build/test suite passes
- [ ] `pyfiction-docstring-generator` workflow regenerates `pybind11_mkdoc_docstrings.hpp` on push
- [ ] Confirm CMakeLists version bump to 0.7.1 happens as part of the normal release process before this ships

## Summary by CodeRabbit

* **Deprecations**
  * Deprecated the `onepass` command; use `exact` instead. Removal is planned for version 0.8.0.
  * Deprecated energy commands and energy-dissipation APIs; removal is planned for version 0.8.0.
  * Deprecated legacy path-finding, physical-design, simulation, and range APIs.
  * Deprecated the `EXGS` simulation mode; use `QUICKEXACT` instead.
  * Added migration guidance for `a_star()`, `exact_physical_design()`, and `quickexact()`.
  * Deprecation warnings now appear during compilation when affected APIs are used.
  * Updated command-line and Python API documentation with the deprecation guidance.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -82,6 +82,6 @@ SpacesInCStyleCastParentheses: false
 SpacesInContainerLiterals: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
-Standard: c++17
+Standard: c++20
 UseTab: Never
 LineEnding: LF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -72,7 +72,7 @@
       matchDepNames: [
         "sphinx"
       ],
-      matchCurrentVersion: "==8.1.3",
+      matchCurrentVersion: "8.1.3",
       enabled: false
     },
     {
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -17,6 +17,96 @@
     "pre-commit",
     "pep621",
     "dockerfile",
+    "custom.regex",
+  ],
+  customManagers: [
+    {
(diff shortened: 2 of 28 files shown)


marcelwa:
Abandoning the deprecate-then-remove-later approach in favor of removing these functions outright in #1043 (still v0.x.y, so the extra deprecation-window overhead isn't worth it). `exhaustive_ground_state_simulation()`/ExGS is the one exception — it stays untouched, neither deprecated nor removed, since it's too deeply embedded to safely remove right now.

Issue #1062, opened 2026-08-14
🐛 Docstring generator parses the code base as C++11 with no include path


marcelwa:
## Summary

The pyfiction docstring generator parses the entire code base as **C++11 with no include path**. This silently drops Doxygen comments, and it is the real reason three separate attempts to replace `static_assert` type checks with `requires` clauses had to be reverted.

Fixing it needs three changes made together — an include path, a pinned libclang, and the standard flag. `-std=c++20` on its own makes things **worse**; that was tried in #1061 and reverted.

## The two defects

`.github/workflows/pyfiction-docstring-generator.yml` invokes:

```
python3 -m pybind11_mkdoc -o ...docstrings.hpp -D FICTION_Z3_SOLVER -D FICTION_ALGLIB_ENABLED `find include/fiction -name "*.hpp" -print`
```

**1. No `-std`, so it parses as C++11.** `read_args()` in `pybind11_mkdoc/mkdoc_lib.py` does:

```python
if not any(it.startswith("-std=") for it in args):
    parameters.append('-std=c++11')
```

A `requires` clause is a syntax error under C++11. It aborts the enclosing declaration list, so every declaration after it loses its Doxygen comment — silently, with a zero exit code. Minimal reproduction, same `clang==15.0.7` pin throughout:

| Header | Flags | Result |
| --- | --- | --- |
| `static_assert` | none (current) | 3 symbols, all documented |
| `requires` | none (current) | **2 symbols lost**, `error: expected ';' at end of declaration list` |
| `requires` | `-std=c++20` | 3 symbols, clean parse |

**2. No `-I include`, so no project include resolves.** Every `#include "fiction/..."` fails. A CI run logs ~130–180 such errors:

```
error: 'fiction/traits.hpp' file not found          (23x)
error: 'fiction/layouts/coordinates.hpp' file not found   (7x)
```

The generator only works today because libclang's error recovery is good enough under C++11. That is the load-bearing part, and it is why the naive fix fails.

## Why `-std=c++20` alone is not the fix

Tried in #1061 (`451e7b2c`), reverted in `f0ddb7a0`:

| Environment | libclang | Docstrings | Unresolved `DOC()` refs |
| --- | --- | --- | --- |
| Local | 22.1.8 | 3283 → **3294** | 4 of 876 (pre-existing) |
| **CI** | runner default | 3292 → **1769** | **422 of 876** |

[...]

PR #1081, opened 2026-08-15
⚡️ Cut compile time, memory, and the CI fleet

marcelwa:
## Description

CI runs regularly exceed two hours, and Debug compiles peak near 4 GB per process. The long-term fix — consolidating layout and network types to get rid of templating — is a rewrite the repo is not ready for. This collects what is available today: include hygiene on the hot path, one template parameter that was carried through ~1900 lines of code that never used it, and a CI topology that built the same thing many times over.

No public API changes.

**On the single-PR shape.** `AGENTS.md` asks for the smallest change and separate PRs for unrelated work. That is deliberately set aside here: every one of these pieces needs a full CI run to validate, and full CI runs are the thing being fixed. Six stacked PRs would mean six serialized multi-hour validation cycles before any of the benefit lands. The commits are separate and reviewable in order.

### How the numbers were measured

Wall time is not reproducible on the measuring host — the governor is `powersave` and core clocks swing between 1.1 and 3.3 GHz, so two identical builds minutes apart differed by 24%. An early measurement showed the `nlohmann/json` removal making the build *slower*, which was pure thermal drift. Everything below is therefore **retired user-space instructions** (`perf stat`) and **peak RSS**, both clock-independent, measured per translation unit single-threaded. GCC 16.1.1, Debug, Z3 + ALGLIB, ccache off.

The eight representative translation units are the two physical-design test TUs, the two CLI physical-design TUs, and four heavy test TUs (`gate_level_layout`, `charge_distribution_surface`, `quickexact`, `dot_drawers`).

### 1. `FICTION_ENABLE_TIME_TRACE`

There was no way to attribute build time to headers or template instantiations, so every attempt at speeding the build up was guesswork. New Clang-only option, off by default, plus ClangBuildAnalyzer instructions in the docs.

### 2. CI topology

Measured on this branch's own green run against `main`'s last green run, summing billable
job minutes:

| workflow | main | this branch |
|---|---|---|
| 🐧 CI | 16 jobs, **815 min** | 9 jobs, **373 min** |
| 🪟 CI | 4 jobs, **293 min** | 2 jobs, **141 min** |
| 🍎 CI | 2 jobs, **88 min** | 1 job, **20 min** |
| 🐍 CI | 4 jobs, **60 min** | removed |
| 🐍 Packaging | 5 jobs, 137 min | 5 jobs, 141 min |
| everything else | 4 jobs, 50 min | 5 jobs, 62 min |
| **total** | **36 jobs, 1443 min** | **23 jobs, 737 min** |

**Roughly half the billable minutes, and a third fewer jobs.** The comparison understates
the steady-state gain if anything: this branch edits headers that nearly everything
includes, so its ccache hit rate was far worse than a typical push to `main`.

Also: experiment binaries compiled per 🐧 run drop from ~1090 to 34.

- The compiler matrix carried g++-11 through 14 and clang++-16 through 20. It now carries the ends of each family plus one point between, with `ubuntu-22.04 × g++-11` as the oldest-supported canary. `docs/getting_started.rst` records the verified set, which it previously did not.
- Experiments are 34 stand-alone paper reproductions that CI never runs, and `ci-common` built all of them in every OS job in both build types. One dedicated 🐧 job now compiles them so they cannot rot.
- 🐍 CI ran `nox -s tests` across five CPythons on four runners — 20 full extension builds — to test what 🐍 Packaging already builds, while the wheels only smoke-tested `from mnt import pyfiction`. The workflow is deleted and the wheels now run the real suite against the repaired wheel via cibuildwheel's `test-sources`/`test-groups`/`test-command`, which is also stronger validation. `nox -s tests` stays as the local entry point.
- ccache was disabled on every `ubuntu-24.04-arm` job over a stale "not supported on ARM yet" comment, leaving the slowest runners permanently cold.
- `FICTION_LIGHTWEIGHT_DEBUG_BUILDS` moves from a hand-passed `-D` in `ubuntu.yml` into the `ci-debug` preset, so macOS, Windows, and CodeQL Debug builds get it too — see the measurement below for why that matters.
- Every build job gets a `timeout-minutes`; none had one, so a hang burned GitHub's six-hour default.
- Drive-bys: the CodeQL ccache key interpolated an undefined `matrix.os`, and 🐍 Packaging's path filter still pointed at `bindings/pyfiction/**`, which has not existed since the bindings were restructured.

### 3. Include hygiene

`gate_level_layout.hpp` included `algorithms/verification/design_rule_violations.hpp` for exactly one line — `friend class detail::gate_level_drvs_impl` — dragging `nlohmann/json.hpp` (~25k lines) plus three `fmt` headers into 28 of the 145 headers under `include/fiction`, and so into effectively every test, CLI, and bindings TU. A forward declaration is all the friend declaration needs; every caller of `gate_level_drvs` already includes the header directly.

`math_utils.hpp` included the vendored `combinations.h` for a single function that four call sites use. Because it sits on the `coordinates.hpp` → `traits.hpp` path, that header reached 113 of 145 fiction headers. The function moved to `fiction/utils/combination_utils.hpp`.

| translation unit | Ginstr | peak RSS |
|---|---|---|
| `test/layouts/gate_level_layout.cpp` | 126.4 → 123.1 | 840 → 770 MiB |
| `test/io/dot_drawers.cpp` | 133.9 → 131.1 | 1124 → 1002 MiB |
| `test/technology/charge_distribution_surface.cpp` | 253.3 → 249.8 | 1394 → 1323 MiB |
| `test/.../quickexact.cpp` | 237.6 → 234.3 | 1477 → 1418 MiB |

TUs that need the design-rule checker anyway are unchanged, as expected. The gain is modest in instructions (~2%) and more meaningful in memory (5–11%).

### 4. `Ntk` dropped from two physical-design implementations

`orthogonal_impl` and `graph_oriented_layout_design_impl` convert the specification network to a `technology_network` in their constructor and then never look at the caller's type again — the members are a fixed `topo_view<fanout_view<names_view<technology_network>>>` and a `tec_nt`. `Ntk` appeared exactly twice in each class, in the template header and the constructor parameter, while 210 and 1690 lines of body were instantiated once per network type in use.

`exact` already had this right (`exact_impl<Lyt>`); both now follow that shape. Public entry points keep their `<Lyt, Ntk>` signatures, so no caller changes. The CLI feels it most, dispatching over a four-way `logic_network_t` variant: `ortho.cpp` goes from 20 instantiations to 5, `gold.cpp` from 4 to 1.

| translation unit | Ginstr | peak RSS |
|---|---|---|
| `test/.../graph_oriented_layout_design.cpp` | 219.5 → 191.7 (**−13%**) | 1508 → 1350 MiB |
| `cli/cmd/physical_design/src/gold.cpp` | 173.5 → 159.6 (−8%) | 1756 → 1684 MiB |
| `cli/cmd/physical_design/src/ortho.cpp` | 160.3 → 150.3 (−6%) | 1660 → 1601 MiB |
| `test/.../orthogonal.cpp` | 220.4 → 213.0 (−3%) | 1551 → 1509 MiB |

`on_the_fly_sidb_circuit_design` keeps its `Ntk`: it forwards straight into `exact_with_blacklist`, already layout-only inside.

### 5. Build knobs

**`FICTION_LIGHTWEIGHT_DEBUG_BUILDS` is by far the biggest single lever**, and before this PR only 🐧 used it:

| config | 8 TUs, Ginstr | worst-TU peak RSS |
|---|---|---|
| lightweight OFF | 1825.7 | 2139 MiB |
| lightweight ON | 1098.7 (**−40%**) | 1509 MiB (**−29%**) |
| lightweight ON + PCH ON | 1050.9 (−42%) | 1508 MiB |

Per-TU memory under the flag: `charge_distribution_surface` 1304 → 829 MiB, `quickexact` 1414 → 1038 MiB, `gate_level_layout` 767 → 560 MiB.

**PCH now covers the test suite.** It previously only covered the CLI's 37 TUs while the 118 test executables each recompiled the same stack; 62 of them include `fiction/types.hpp`, which instantiates every preset layout type. They share one PCH via a carrier target. Worth 4–8% on test TUs on top of the lightweight flag. `Cache.cmake` sets `CCACHE_SLOPPINESS=pch_defines,time_macros` when PCH is on — without it ccache refuses to cache PCH-using compilations, which would make PCH a net loss. Verified 100% ccache hit rate with both on. `dev` and `tests-slim` enable PCH and the lightweight flag; CI presets leave PCH off, where ccache dominates and PCH costs a little memory.

**`FICTION_ENABLE_UNITY_BUILD` is removed, not fixed.** It set `UNITY_BUILD` on an `INTERFACE` target, where the property does not propagate, so it has never done anything. Unity builds also raise peak memory, which is the constraint here.

### 6. A test that had never run

`bdl_input_iterator.py` did not match pytest's default `python_files = test_*.py`, so its five tests were never collected. Renamed; all five pass. 179 → 184 collected.

## Verification

- Full Debug build of everything CI builds (118 test executables + 34 experiments + CLI, Z3 + ALGLIB), **GCC 16.1.1 and Clang 22.1.8**: both clean.
- `ctest`: 118/118 under GCC, 117/117 under Clang.
- ccache hit rate with PCH enabled: 1/1 (100%).
- The `FORCE` fix on the compiler launcher was verified by configuring a build tree with PCH off, then reconfiguring the same tree with PCH on and confirming the launcher picks up the `CCACHE_SLOPPINESS` wrapper.
- **CI is green**: 29 checks passing, 0 failing.
- The wheel builds really do run the suite now, not a no-op: the macOS wheel job's log shows `185 passed` six times over, once per CPython, against the repaired wheels.

## Notes for the reviewer

- **Branch protection**: 🐍 CI is deleted, so its required-check entry needs removing from the branch protection rules or PRs will block on a check that can never report.
- The `.clang-format` UTF-8 BOM that made `prek run -a`'s `check-yaml` hook fail on `main` is fixed here too, since it blocked running the hooks locally at all.
- Worth a separate PR, not touched here: `test/io/network_reader.cpp` hard-codes `../../benchmarks/TOY/mux21.v`, so it only passes when the build directory sits inside the source tree.
- The docstring generator pushes a commit to the PR branch on every `.hpp` change, and the workflows for that bot commit land in `action_required` and need approving by hand before they run. Not introduced here, but it is a manual step on every push that touches headers.
- Still on the table for later, with `-ftime-trace` now available to justify it: splitting `types.hpp` so a TU needing `gate_layout_t` does not instantiate ~30 SiDB layout types (~106 include sites), and converting the 77 `void_t` detection idioms in `traits.hpp` to C++20 `requires` — the latter blocked on #1062, since bare `requires` clauses break the docstring generator.

## Checklist:

- [x] I have created/adjusted the Python bindings for any new or updated functionality. (No binding surface changed.)


marcelwa:
Thanks — four of the five are in, one I'm declining.

**`cmake/Cache.cmake` — fixed.** Real bug, and a good catch: a build tree configured before `FICTION_ENABLE_PCH` was turned on kept its bare `ccache` launcher, so the `CCACHE_SLOPPINESS` wrapper never applied and every PCH compilation silently missed the cache. Both assignments now carry `FORCE`, with a comment saying why.

**`docs/changelog.rst` — fixed.** The entries were well over the one-or-two-sentence, ~90-character profile that `docs/AGENTS.md` asks for, and several were per-file refactor enumerations that the same file explicitly rules out. Rewritten; the detail lives in the PR description and in code comments at the sites. Also reordered the `Added` categories alphabetically, which I had broken.

**`docs/getting_started.rst` — fixed.** You're right that the prose implied every compiler runs on both Ubuntu architectures. Replaced with a table listing each platform against the compilers actually built there.

**`orthogonal.hpp` — fixed.** Added a Doxygen block to the constructor and a `@tparam` to the class, matching what `graph_oriented_layout_design_impl` already carries.

**`combination_utils.hpp` — declining.** Re-exposing `combination_utils.hpp` from `math_utils.hpp` would undo the change. `math_utils.hpp` sits on the `coordinates.hpp` → `traits.hpp` path, so anything it includes reaches 113 of the 145 headers under `include/fiction`; the entire point of the move was to get the vendored `combinations.h` off that path for the one function in 145 headers that uses it. Adding the include back restores exactly the cost the commit removes.

The break is real, so it is now marked **Breaking:** in the changelog with the header to include instead, rather than papered over.

PR #846, opened 2025-09-29
:sparkles: Add molecular QCA gate library

hibenj:
## Description

Add molecular QCA as a technology for cartesian layouts. In this PR the library is added with the corresponding changes for printing `.svg`s, generate `.qll`-files and adding the command to the cli.

Fixes # (issue)

## Checklist:

- [x] The pull request only contains commits that are related to it.
- [x] I have added appropriate tests and documentation.
- [ ] I have added a changelog entry.
- [ ] I have created/adjusted the Python bindings for any new or updated functionality.
- [ ] I have made sure that all CI jobs on GitHub pass.
- [ ] The pull request introduces no new warnings and follows the project's style guidelines.

## Summary by CodeRabbit

* **New Features**
  * Added MolQCA (molecular QCA) support: new gate library, MolQCA layout type, SVG and QLL export paths, CLI handling, and a technology trait.

* **Documentation**
  * Large expansion of Python docstrings and API docs for layout export, rendering, pathfinding, analysis and utilities; added publication entry.

* **Refactor**
  * Standardized nested enum scoping, improved type-dispatch logic and trait checks, safer optional handling.

* **Tests**
  * Extensive tests for MolQCA library, layouts, blueprints and related flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cda-tum/fiction/pull/846?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

Issue #1091, opened 2026-08-17
🔧 Decide whether the project adopts per-file copyright headers (`CPY001`)


marcelwa:
`[tool.ruff.lint] ignore` carried a block headed `# TODO checks to be added back in and
addressed:`. #1090 clears it. `CPY001` (missing-copyright-notice) is the one entry that does not
belong there, so it is split out here and stays ignored with a comment recording the reason.

`CPY001` fires on all 87 tracked Python files, because the project uses no per-file copyright
headers anywhere. Enabling it is not cleanup — it means adopting a header on every Python file
and deciding what it says, who it names, and whether the C++ tree follows.

Three ways this can close:

- **Won't fix.** The `LICENSE.txt` at the repository root and the license metadata in
  `pyproject.toml` already state the terms. The ignore entry stays as it is now, with the reason
  recorded.
- **Adopt headers for Python only.** 87 files, mechanical once the text is fixed.
- **Adopt headers repository-wide.** Then the C++ tree needs the same treatment, and a
  `clang-format`- or `prek`-level check to keep new files in line.

Nothing depends on this, and the ignore entry now names this issue, so the TODO block is gone
either way.

Split out of #1090.

---

Filed with AI assistance; the file count was measured with `ruff` 0.16.3.

PR #1136, opened 2026-08-19
🐛 Fix `noexcept`, SVG double-hash, and `qll` CLI description

### LINKED ISSUE #1129: Fix noexcept on determine_port_routing in gate libraries
## Summary

`determine_port_routing` in `molecular_qca_library` (and the analogous function in `qca_one_library`) is marked `noexcept` while its body calls `p.inp.emplace` / `p.out.emplace`, which can throw `std::bad_alloc`. An allocation failure therefore terminates rather than propagating.

## Affected files

- `include/fiction/technology/molecular_qca_library.hpp` — `determine_port_routing` (line ~139)
- `include/fiction/technology/qca_one_library.hpp` — analogous function

## Suggested fix

Remove `noexcept` from both functions, or replace `emplace` with a non-throwing alternative if the sets are pre-allocated.

Fix both together to avoid a divergence between the two libraries.

## Context

Flagged in PR #846 (Add molecular QCA library). Deferred from that PR to avoid introducing an inconsistency between the two libraries mid-review.

hibenj:
## Description

Follow-up fixes flagged during review of #846.

- **Remove `noexcept` from all six `determine_port_routing` overloads** (`qca_one_library`, `sim7_mol_library`, `inml_topolinano_library`, `sidb_bestagon_library`, `sidb_on_the_fly_gate_library`, `sidb_skeleton_bestagon_library`): every overload emplaces into a `port_list`'s `std::set` members, which can throw `std::bad_alloc`. An allocation failure would previously terminate rather than propagate. The sole caller in each case, `set_up_gate`, is already non-`noexcept` and only catches `std::out_of_range`, so `std::bad_alloc` now propagates cleanly without further changes.

- **Fix `fill:##000000` double-hash bug** in `write_qca_layout_svg_impl`: `cell_color` was set to `"#000000"` for constant-0/1 cells. That value is only ever consumed by `SIMPLE_CELL`'s `fill:#{0}` format string, which the non-simple-mode `CONST0`/`CONST1` templates don't reference — so the malformed `fill:##000000` could only reach output when `ps.simple == true`. Fixed to use the bare `"000000"` form already used by `sim7_mol_library`'s SVG writer, and factored into a shared `svg::CONST_CELL` constant (alongside `PI_CELL`/`PO_CELL`) used at all four QCA/molQCA call sites instead of a literal duplicated four times. Added a regression test, since `write_qca_layout_svg_impl` had zero coverage before this PR.

- **Update `qll` CLI description**: the help text still said "iNML or QCA" even though molQCA support was added in #846. Updated `qll.cpp`'s description to mention molQCA and SCERPA, and brought `qll.hpp`'s Doxygen comments (class doc, link, `execute()` doc) in line with it and with `docs/cli.rst`.

Added changelog entries under `Unreleased -> Fixed` for the two user-facing changes (SVG export, `qll` help text).

Closes #1129

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * QLL output documentation now includes support for molQCA layouts and SCERPA simulations, alongside existing formats and tools.

* **Bug Fixes**
  * Corrected SVG rendering for constant-0 and constant-1 QCA cells in simple mode, ensuring the intended black fill is displayed.

* **Documentation**
  * Updated changelog and command descriptions to reflect expanded layout and simulation support.


include/fiction/technology/qca_one_library.hpp
```diff
@@ -146,7 +146,7 @@ class qca_one_library : public fcn_gate_library<qca_technology, 5, 5>
 
   private:
     template <typename Lyt>
-    [[nodiscard]] static port_list<port_position> determine_port_routing(const Lyt& lyt, const tile<Lyt>& t) noexcept
+    [[nodiscard]] static port_list<port_position> determine_port_routing(const Lyt& lyt, const tile<Lyt>& t)
```
marcelwa, line 149: `port_list::inp` and `out` are `std::set` (`cell_ports.hpp:208`), so every `emplace` here can throw `std::bad_alloc`. That reasoning is sound — but it applies verbatim to four more overloads that this PR leaves `noexcept`:

| File | Line |
| --- | --- |
| `include/fiction/technology/inml_topolinano_library.hpp` | 383 |
| `include/fiction/technology/sidb_bestagon_library.hpp` | 306 |
| `include/fiction/technology/sidb_on_the_fly_gate_library.hpp` | 848 |
| `include/fiction/technology/sidb_skeleton_bestagon_library.hpp` | 112 |

All four `emplace` into the same `std::set` members while declared `noexcept`. The PR description says *"Fixes both libraries together to avoid divergence"*, but two-of-six **creates** divergence: two libraries now propagate an allocation failure, four still terminate on it. #1129 named only two because #846's review only touched those two; its title says "in gate libraries", plural.

Please drop `noexcept` from all six so the codebase has one consistent answer. In every case the sole caller is `set_up_gate`, which is already non-`noexcept` and throws `unsupported_gate_type_exception` / `unsupported_gate_orientation_exception`, so nothing downstream needs to change and the propagation path already exists.

Worth double-checking `sidb_bestagon_library.hpp:83` and `sidb_on_the_fly_gate_library.hpp:226` while you're in there — both call `determine_port_routing` from inside a loop over adjacent tiles, so confirm the surrounding code is exception-safe once the call can throw.

include/fiction/technology/sim7_mol_library.hpp
```diff
@@ -136,7 +136,7 @@ class sim7_mol_library : public fcn_gate_library<mol_qca_technology, 10, 10>
      * @return Incoming and outgoing molQCA connector positions for `t`.
      */
     template <typename Lyt, bool RespectClocking = true>
-    [[nodiscard]] static port_list<port_position> determine_port_routing(const Lyt& lyt, const tile<Lyt>& t) noexcept
+    [[nodiscard]] static port_list<port_position> determine_port_routing(const Lyt& lyt, const tile<Lyt>& t)
```
marcelwa, line 139: Same change here, and correct for the same reason. Please extend it to the remaining four overloads listed in my comment on `qca_one_library.hpp` so all six agree.

One thing to verify while you're here: unlike `qca_one_library`, this library also calls `determine_port_routing<GateLyt, false>` from *inside* the `try` block at line 70. `std::bad_alloc` won't be caught by the `catch (const std::out_of_range&)` handler, so it correctly propagates rather than being misreported as an unsupported gate orientation — but it's worth a conscious check rather than an accident.

include/fiction/io/write_svg_layout.hpp
```diff
@@ -960,12 +960,12 @@ class write_qca_layout_svg_impl
         }
         else if (Lyt::technology::is_const_0_cell(ct))
         {
-            cell_color       = "#000000";
+            cell_color       = "000000";
```
marcelwa: The fix is right, but the bug is narrower than the PR description implies, and the difference matters for the test.

`CONST0` and `CONST1` never reference `{0}` — they only use `{1}`/`{2}` for the `translate`. So in non-`simple` mode `cell_color` was computed and discarded, and this change is a no-op there. The malformed `fill:##000000` could **only** ever reach the output when `ps.simple == true`, via `SIMPLE_CELL`, which does use `fill:#{0}`. Please correct the description so the actual blast radius is on record.

That also pins down what the regression test needs. There isn't one, and there's nothing to fall back on: `write_qca_layout_svg_impl` has **zero** coverage today. Every `TEMPLATE_TEST_CASE` in `test/io/write_svg_layout.cpp` is SiDB or CDS, and `bindings/mnt/pyfiction/test/inout/test_write_svg_layout.py` is the same. `codecov/patch` reporting green on this PR is therefore not evidence the fix works — worth a glance at why it passed.

A single QCA layout with one `CONST_0` cell, written with `ps.simple = true`, asserting the output contains `fill:#000000` and no `##`, is about ten lines and closes the gap permanently.

include/fiction/io/write_svg_layout.hpp
```diff
@@ -960,12 +960,12 @@ class write_qca_layout_svg_impl
...
             cell_description = ps.simple ? svg::SIMPLE_CELL : svg::CONST0;
         }
         else if (Lyt::technology::is_const_1_cell(ct))
         {
-            cell_color       = "#000000";
+            cell_color       = "000000";
```
marcelwa: Every other color in this writer comes from a named `svg::` constant — `PI_CELL`, `PO_CELL`, `CLOCK_ZONE_*`. This fix leaves a bare `"000000"` duplicated at lines 963 and 968, and `write_mol_qca_layout_svg_impl` has the identical pair at 1264 and 1269. Four copies of a magic literal whose correctness depends on an unwritten "no leading hash here" rule is precisely how this bug happened in the first place.

Please add a constant alongside `PI_CELL`/`PO_CELL` (around line 202):

```cpp
inline constexpr const char* CONST_CELL = "000000";
```

and use `svg::CONST_CELL` at all four sites. That makes the convention explicit at the point of definition rather than implicit at each use, and it's the smallest step toward #1138.

cli/cmd/io/src/qll.cpp
```diff
@@ -25,8 +25,9 @@ namespace alice
 {
 
 qll_command::qll_command(const environment::ptr& e) :
-        command(e, "Generates a ToPoliNano/MagCAD layout file for the current iNML or QCA cell layout in store. "
-                   "ToPoliNano can reuse the layout while MagCAD is able to perform physical simulations.")
+        command(e, "Generates a ToPoliNano/MagCAD/SCERPA layout file for the current iNML, QCA, or molQCA cell "
```
marcelwa, line 28: The help text update is accurate — I verified `qll.cpp:52` guards on `has_mol_qca_technology_v` and `cell.cpp` registers the `SIM7-MOL` library, so molQCA layouts really are reachable here.

But the class documentation two files over wasn't updated with it and is now the only stale spot:

- `cli/cmd/io/include/qll.hpp:16` — "Generates a ToPoliNano/MagCAD layout file for the current cell layout in store..."
- `cli/cmd/io/include/qll.hpp:18` — links only `https://topolinano.polito.it/`
- `cli/cmd/io/include/qll.hpp:32` — "Function to perform the output call. Generates a ToPoliNano/MagCAD layout file."

`docs/cli.rst:324` already got this right in #846 and links SCERPA as `https://ieeexplore.ieee.org/document/8935211`; please bring the header in line with both.

marcelwa (changes requested):
Thanks for the follow-up — all three changes are individually correct, and CI is green. Requesting changes on **scope and completeness** rather than correctness: as it stands the `noexcept` fix leaves the inconsistency it set out to remove, and two of the three changes are user-facing with no changelog entry.

Summary of what I'd like before merge:

1. **Streamline `noexcept` across all six `determine_port_routing` overloads**, not two. Details inline on `qca_one_library.hpp`.
2. **Add a changelog entry.** `AGENTS.md` (Boundaries -> Always) requires `docs/changelog.rst`'s `Unreleased` section to be updated for any user-facing change. Two of the three changes qualify: the SVG writer emitted malformed markup that reached users, and the `qll` help text is user-visible. `Unreleased -> Fixed` already has the buckets you need. Please also fill in `.github/pull_request_template.md`'s checklist in the PR body — it's the other AGENTS.md "always", and it would have caught both this and the missing test.
3. **Add a regression test for the SVG fix**, and **update the stale `qll.hpp` Doxygen comment** — details inline.

I filed #1138 for the underlying cause of the SVG bug (two clashing color-constant conventions in one file, plus the QCA/molQCA writer duplication). That's #846's debt, not this PR's — no action needed here, but it explains why this class of bug recurs and why the regression test below is worth the ten lines.

Minor, non-blocking: the single commit bundles three unrelated fixes. Grouping them into one PR is the right call given the ~2h CI gate, but three commits would review better and would match the template's "the pull request only contains commits that are related to it."

hibenj:
Addressed all the review feedback:

1. **noexcept streamlined across all six overloads** — removed from `inml_topolinano_library`, `sidb_bestagon_library`, `sidb_on_the_fly_gate_library`, and `sidb_skeleton_bestagon_library` in addition to the two already fixed. Checked both call sites flagged (`sidb_bestagon_library.hpp:83`, `sidb_on_the_fly_gate_library.hpp:226`): both `catch` blocks only catch `std::out_of_range`, so `std::bad_alloc` propagates cleanly regardless of whether the call is inside or outside the `try` block.
2. **Changelog entries added** under `Unreleased -> Fixed` for the SVG fix and the `qll` help-text update, and the PR checklist is filled in.
3. **Regression test added** for the SVG fix (`write_qca_layout_svg_impl` had none before), and `qll.hpp`'s stale Doxygen comments are now in line with `qll.cpp` and `docs/cli.rst`.
4. **`svg::CONST_CELL` constant added**, replacing all four bare `"000000"` literals.
5. **PR description corrected** to note the double-hash bug was only reachable in simple mode (`ps.simple == true`), since non-simple `CONST0`/`CONST1` templates never reference `cell_color`.
6. **Split into 4 commits** on top of the original one: noexcept streamlining, SVG fix + test, `qll.hpp` doc fix, and a small clang-format fixup.

docs/changelog.rst
```diff
@@ -174,6 +174,12 @@ Fixed
       traced contour came out empty and every reachable point was marked operational without simulation
     - Fixed a data race on ``quicksim``'s timeout flag, which every worker thread wrote as a plain
       ``bool``. It is now ``std::atomic_bool``
+- I/O:
+    - Fixed ``write_qca_layout_svg``'s simple-mode output emitting malformed ``fill:##000000`` for
+      constant-0/1 QCA cells instead of ``fill:#000000``
+- CLI:
+    - Updated the ``qll`` command's description to mention the molQCA layouts and SCERPA simulations
```
marcelwa: Two small things:

- Drop the `#846` reference. The changelog is read by users, to whom a PR number means nothing, and #846's own entry is a few lines up in the same `Unreleased` block under `Added -> Gate libraries`, so the pointer is redundant even for us. `"to mention molQCA layouts and SCERPA simulations"` says everything the reader needs. (The one existing `#`-reference in this file, `#1091` on line 128, is a live forward pointer to something still open — different case.)
- Swap the `I/O` and `CLI` blocks so `CLI` comes first, matching the relative order in the `Added` section above.

test/io/write_svg_layout.cpp
```diff
@@ -959,6 +959,23 @@ TEST_CASE("Generate MolQCA layout in SVG", "[write-mol-qca-layout-svg]")
     }
 }
 
+TEST_CASE("Generate QCA layout in simple SVG mode with constant cells", "[write-qca-layout-svg]")
```
marcelwa, line 969: The test itself is right — it's placed in the wrong spot. It splits the three `[write-mol-qca-layout-svg]` cases (lines 895, 908, 979) down the middle. Please move it to after the `"Reject invalid MolQCA SVG output files"` block so the molQCA group stays contiguous and this becomes the start of the QCA group.

include/fiction/io/write_svg_layout.hpp
```diff
@@ -960,12 +961,12 @@ class write_qca_layout_svg_impl
         }
         else if (Lyt::technology::is_const_0_cell(ct))
         {
-            cell_color       = "#000000";
+            cell_color       = svg::CONST_CELL;
             cell_description = ps.simple ? svg::SIMPLE_CELL : svg::CONST0;
```
marcelwa, line 969: Not for this PR — for #1138, whose scope this belongs in.

At line 1148, `ps.simple ? "" : text_colors[czone]` feeds `TILE`'s `fill:#{4}`, so every tile in simple mode renders as `style="...fill:#;stroke:none;"`. That is the same class of malformed-color bug you're fixing here, it's pre-existing, and the SVG your new test generates contains it. Worth noting that `CHECK_FALSE(contains(svg, "fill:##"))` won't catch this shape of it — a guard on `fill:#;` would, but the real fix is to omit the whole `<text>` element in simple mode rather than to emit it with an empty color. Please add it to #1138.

include/fiction/technology/inml_topolinano_library.hpp
```diff
@@ -212,7 +213,8 @@ class inml_topolinano_library : public fcn_gate_library<inml_technology, 4, 4>
             }
         };
 
-        do
+        do  // NOLINT(cppcoreguidelines-avoid-do-while): the loop body must run at least once to
```
marcelwa: AGENTS.md asks a surviving `NOLINT` to name the *technical* reason it has to stay. "the loop body must run at least once" restates what `do`/`while` means; it doesn't say why the alternative is worse. What actually justifies it here is that the loop's exit condition depends on state the body computes, so hoisting it into a `while` would mean duplicating the body or introducing a flag variable. If you reword it to say that, the suppression documents a decision instead of a definition.

marcelwa (changes requested):
Many thanks for working through all of this — every point from the last round is addressed, and I verified the two things I had asked you to double-check rather than change; both hold up. We're almost there. Just a handful of small follow-ups inline, none of them structural.

Please also tick the "all CI jobs pass" box in the checklist; all 25 jobs are green on `84dc6c23`.

hibenj:
Addressed:

- Checklist: ticked "all CI jobs pass".
- Changelog: dropped the `#846` reference, swapped CLI/I/O order to match the Added section.
- Test placement: moved the QCA test after "Reject invalid MolQCA SVG output files" so the molQCA group stays contiguous.
- `TILE`'s `fill:#{4}` bug in simple mode: filed as a comment on #1138 (same underlying cause, out of scope here).
- NOLINT reworded to state the actual reason (exit condition computed by the loop body, so a `while` would need to duplicate the body or add a flag variable).

marcelwa:
Looking good. Please resolve the merge conflicts for the CI to run.

marcelwa (approved):
Very good, many thanks for the fix and iteration! 🙏 In it goes 😄

PR #1137, opened 2026-08-20
♻️ Migrate FCN cell_type enums to scoped enum class

### LINKED ISSUE #1060: Migrate FCN technology cell_type enums to scoped enums
## Context

PR #846 adds temporary clang-tidy suppressions for the existing unscoped `cell_type` enums in FCN technology definitions because adding `mol_qca_technology` makes `cell_technologies.hpp` part of the changed-file clang-tidy surface.

## Follow-up

Migrate the public `cell_type` enums for existing FCN technologies to scoped enums in a dedicated PR.

This should include:

- a changelog entry,
- updates to downstream call sites,
- updates to affected tests,
- clear migration visibility for downstream users.

## Notes

This is intentionally split out of #846 to keep that PR focused on the MolQCA technology, gate library, writers, and bindings.

hibenj:
## Summary

Converts the `cell_type` enums in `qca_technology`, `inml_technology`, and `sidb_technology` from unscoped `enum` to `enum class`, making them consistent with `mol_qca_technology::cell_type` which was already scoped (introduced in #846).

- Removes the `// NOLINT(cppcoreguidelines-use-enum-class)` suppressions that were added as temporary workarounds
- Updates `cell_mark` initializers in all three structs to use `static_cast<uint8_t>(cell_type::XXX)` (required for scoped enums)
- Qualifies all bare enum comparisons inside the struct helpers (e.g. `c == EMPTY` → `c == cell_type::EMPTY`)
- Updates 10 external call sites that used the short unscoped form (e.g. `sidb_technology::INPUT` → `sidb_technology::cell_type::INPUT`) in `is_operational.hpp`, `sidb_on_the_fly_gate_library.hpp`, and `design_sidb_gates.cpp`
- Adds a **Breaking** changelog entry

## Migration

Before:
```cpp
sidb_technology::INPUT
qca_technology::NORMAL
inml_technology::EMPTY
```

After:
```cpp
sidb_technology::cell_type::INPUT
qca_technology::cell_type::NORMAL
inml_technology::cell_type::EMPTY
```

The fully qualified form (e.g. `sidb_technology::cell_type::INPUT`) was already the predominant style across the codebase; this change makes it the only valid one.

Closes #1060

## Summary by CodeRabbit

* **Breaking Changes**
  * Cell types for QCA, iNML, and SiDB technologies are now scoped enumerations.
  * Code using these values must access enumerators with the qualified `cell_type::` prefix.
* **Bug Fixes**
  * Updated SiDB simulation and gate-handling logic to recognize revised cell type values correctly.
  * Improved cell-list decoding with explicit handling for normal, input, output, logic, empty, and default values.
* **Tests**
  * Updated coverage for gate layouts, generated layouts, printed layouts, and bounding boxes.
  * Added validation for optional generated-layout results.
* **Documentation**
  * Reorganized unreleased changelog entries and documented recent build, testing, deployment, and documentation updates.


marcelwa:
@hibenj thanks for getting this started. Once the CI passes and our lovely Mr. Bunny's review comments are addressed, I'll gladly take a look.

hibenj:
🚨 Clang-Tidy is red, but it's pre-existing whole-file debt, not something this PR introduced. The workflow runs clang-tidy over the *entire* file whenever it's in the changed-files list (no `lines-changed-only`), so touching `design_sidb_gates.cpp`/`is_operational.hpp`/etc. surfaces unrelated warnings (`use designated initializer list`, `unchecked access to optional value`, ...) that already existed on `main`. Confirmed: the original commit here already failed the same way before any follow-up fixes.

@marcelwa — is this check required for merge, or fine to leave red here given the unrelated debt? Happy to clean it up if you'd rather have it green.

marcelwa:
> 🚨 Clang-Tidy is red, but it's pre-existing whole-file debt, not something this PR introduced. The workflow runs clang-tidy over the _entire_ file whenever it's in the changed-files list (no `lines-changed-only`), so touching `design_sidb_gates.cpp`/`is_operational.hpp`/etc. surfaces unrelated warnings (`use designated initializer list`, `unchecked access to optional value`, ...) that already existed on `main`. Confirmed: the original commit here already failed the same way before any follow-up fixes.
> 
> @marcelwa — is this check required for merge, or fine to leave red here given the unrelated debt? Happy to clean it up if you'd rather have it green.

Yes, this is intended in an effort to streamline the repo and clear its technical debt piece by piece. Every time someone touches a file, it should be fully cleaned up in the same PR.

test/algorithms/simulation/sidb/random_sidb_layout_generator.cpp
```diff
@@ -28,21 +29,21 @@ TEST_CASE("Random cube::coord_t layout generation", "[random-sidb-layout-generat
         const auto lyt = generate_random_sidb_layout<sidb_cell_clk_lyt_cube>(params);
 
         REQUIRE(lyt.has_value());
-        CHECK(lyt.value().num_cells() == 0);
-        CHECK(lyt.value().x() == 0);
-        CHECK(lyt.value().y() == 0);
+        CHECK(lyt.value().num_cells() == 0);  // NOLINT(bugprone-unchecked-optional-access)
+        CHECK(lyt.value().x() == 0);          // NOLINT(bugprone-unchecked-optional-access)
+        CHECK(lyt.value().y() == 0);          // NOLINT(bugprone-unchecked-optional-access)
```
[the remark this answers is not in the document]
hibenj: You're right about the missing guard — added `REQUIRE(result_lyt.has_value())` at test/algorithms/simulation/sidb/random_sidb_layout_generator.cpp:535 in 0d4bae6d (that call site had none, unlike every other SECTION here).

Kept the NOLINT(bugprone-unchecked-optional-access) comments though: clang-tidy 21 (what CI runs) doesn't recognize Catch2's REQUIRE(...has_value()) as a value-check idiom, so removing them reintroduces the same warning CI already flags as an error on every other `.value()` call in this file — confirmed locally. This matches the existing repo convention in test/algorithms/physical_design/exact.cpp and test/algorithms/simulation/sidb/is_ground_state.cpp, which keep both the guard and the suppression.

include/fiction/technology/sidb_on_the_fly_gate_library.hpp
```diff
@@ -764,27 +764,28 @@ class sidb_on_the_fly_gate_library : public fcn_gate_library<sidb_technology, 60
...
                     }
-                    case 'l':  // output cell
+                    case Lyt::technology::cell_type::LOGIC:  // output cell
                     {
                         lyt.assign_cell_type(all_cell[counter], Lyt::technology::cell_type::LOGIC);
                         break;
                     }
-                    case ' ':
+                    case Lyt::technology::cell_type::EMPTY:
+                    default:
```
marcelwa: This `default:` is new in this PR, and it gives back the exact safety the migration just bought. The switch is now over `cell_type`, all five enumerators are covered, and `-Wall` (`cmake/CompilerWarnings.cmake:47`) gives us `-Wswitch` — but `-Wswitch` goes silent the moment a `default` label is present. Add a sixth SiDB cell type tomorrow and this decoder silently drops it instead of failing the build.

`bugprone-branch-clone` fired in 22e3058e because `case EMPTY: break;` and `default: break;` are identical bodies. The fix is to delete this `default:` line, not to merge the two — with `case Lyt::technology::cell_type::EMPTY:` still listed directly above, the switch is exhaustive and the checker has nothing to clone:

```cpp
                    case Lyt::technology::cell_type::EMPTY:
                    {
                        break;
                    }
```

include/fiction/technology/sidb_on_the_fly_gate_library.hpp
```diff
@@ -764,27 +764,28 @@ class sidb_on_the_fly_gate_library : public fcn_gate_library<sidb_technology, 60
...
                         break;
                     }
-                    case 'o':  // output cell
+                    case Lyt::technology::cell_type::OUTPUT:  // output cell
                     {
                         lyt.assign_cell_type(all_cell[counter], Lyt::technology::cell_type::OUTPUT);
                         break;
                     }
-                    case 'l':  // output cell
+                    case Lyt::technology::cell_type::LOGIC:  // output cell
```
marcelwa: Copy-paste leftover from the `OUTPUT` case above. Pre-existing, but you are rewriting the line anyway.

```suggestion
                    case Lyt::technology::cell_type::LOGIC:  // logic cell
```

test/algorithms/simulation/sidb/random_sidb_layout_generator.cpp
```diff
@@ -28,21 +29,21 @@ TEST_CASE("Random cube::coord_t layout generation", "[random-sidb-layout-generat
         const auto lyt = generate_random_sidb_layout<sidb_cell_clk_lyt_cube>(params);
 
         REQUIRE(lyt.has_value());
-        CHECK(lyt.value().num_cells() == 0);
-        CHECK(lyt.value().x() == 0);
-        CHECK(lyt.value().y() == 0);
+        CHECK(lyt.value().num_cells() == 0);  // NOLINT(bugprone-unchecked-optional-access)
```
marcelwa: 74 bare `// NOLINT(bugprone-unchecked-optional-access)` in one file. The `AGENTS.md` rule is *"a suppression that has to stay names the technical reason in a comment"* — your reasoning (clang-tidy 21 does not model Catch2's `REQUIRE(...has_value())` as a value check) is sound, but it lives in a GitHub thread, not in the code, so the next person to open this file has to go archaeology.

The repo already has the right idiom for exactly this — `NOLINTBEGIN(...)` with the reason after a colon, e.g. `bindings/mnt/pyfiction/src/pyfiction/inout/read_sqd_layout.cpp:55`. Wrapping each `TEST_CASE` records the reason once and, as a bonus, lets you revert the ~20 lines that got reflowed purely to make room for a trailing comment (`can_positive_charges_occur` split across two lines, etc.), which shrinks this diff a lot:

```cpp
// NOLINTBEGIN(bugprone-unchecked-optional-access): clang-tidy 21 does not recognize
// Catch2's REQUIRE(...has_value()) as a value check; every access below is guarded.
TEST_CASE("Random cube::coord_t layout generation", "[random-sidb-layout-generator]")
...
// NOLINTEND(bugprone-unchecked-optional-access)
```

docs/changelog.rst
```diff
@@ -97,6 +97,10 @@ Changed
       ``setuptools`` from the build. Building from source now requires ``scikit-build-core`` 1.0
     - The CMake policy range now ends at 4.4
 - Code quality:
+    - **Breaking:** ``qca_technology::cell_type``, ``inml_technology::cell_type``, and
+      ``sidb_technology::cell_type`` are now scoped ``enum class``, consistent with
+      ``mol_qca_technology::cell_type``. Unqualified access such as ``sidb_technology::INPUT`` must be
+      updated to ``sidb_technology::cell_type::INPUT``
```
marcelwa: Two things. First, CodeRabbit's terminology point is right: `sidb_technology::INPUT` *is* qualified — what changed is that the enumerator no longer leaks into the enclosing struct's scope. "Unqualified access" reads as if `INPUT` alone used to work.

Second, and more useful to a downstream user: scoping also removes the implicit conversion to integer/char, which is the half of the break that will not be obvious from the entry as written. `char c = lyt.get_cell_type(x);` compiled before and does not now.

Suggest replacing the whole entry (lines 100-103) with:

```rst
    - **Breaking:** ``qca_technology::cell_type``, ``inml_technology::cell_type``, and
      ``sidb_technology::cell_type`` are now scoped ``enum class``, consistent with
      ``mol_qca_technology::cell_type``. Replace references to the leaked enumerators, e.g.,
      ``sidb_technology::INPUT``, with ``sidb_technology::cell_type::INPUT``, and add an
      explicit ``static_cast`` wherever a cell type was implicitly converted to an integer
      or ``char``. ``pyfiction`` is unaffected
```

include/fiction/technology/cell_technologies.hpp
```diff
@@ -60,46 +60,94 @@ struct qca_technology
...
         return is_const_0_cell(c) || is_const_1_cell(c);
     }
 
+    /**
+     * Checks whether the given cell mode is the normal mode.
+     *
+     * @param m Cell mode to check.
+     * @return `true` iff `m` is `cell_mode::NORMAL`.
+     */
     [[nodiscard]] static constexpr bool is_normal_cell_mode(const cell_mode& m) noexcept
```
marcelwa, line 151: Optional, your call — I will not block on this one.

The Doxygen you added satisfies `AGENTS.md` (docs on *edited* code) and closes CodeRabbit's request. It does leave the file in a patchwork state though: `qca_technology::is_normal_cell_mode` here is documented while its three siblings at :156, :161, :166 are not, and all twelve `mol_qca_technology` predicates (:263-:360) plus `inml`/`sidb` `is_normal_cell_mode` (:559, :688) are still bare.

Not required by the rules, but the blocks are two lines each and identical in shape to what you have already written, so finishing the file now is cheaper than a follow-up PR that touches this header again and drags in its whole Clang-Tidy backlog.

marcelwa (changes requested):
Thank you for the follow-up! Just a few comments below that should be quick to resolve.

include/fiction/technology/sidb_on_the_fly_gate_library.hpp
hibenj: Removed the `default:`, kept only the explicit `EMPTY` case. Good catch — that's the whole point of the migration. Fixed in 14a30d3d.

include/fiction/technology/sidb_on_the_fly_gate_library.hpp
hibenj: Fixed the comment to "// logic cell" in 14a30d3d.

test/algorithms/simulation/sidb/random_sidb_layout_generator.cpp
hibenj: Switched to one NOLINTBEGIN/NOLINTEND(bugprone-unchecked-optional-access) pair per TEST_CASE with the reason inline, matching read_sqd_layout.cpp:55, and reverted the lines that were only reflowed to fit the trailing comments. Diff for that file is a lot smaller now. Fixed in 14a30d3d.

docs/changelog.rst
hibenj: Took your suggested wording verbatim — good catch on both points (qualified-vs-leaked, and the implicit conversion loss). Fixed in 14a30d3d.

include/fiction/technology/cell_technologies.hpp
hibenj: Finished documenting the rest: qca's remaining cell_mode predicates, all twelve mol_qca predicates, and inml/sidb's is_normal_cell_mode. File is fully documented now. Fixed in 14a30d3d.

marcelwa (approved):
Thank you for the follow-up PR! 🙏

PR #1145, opened 2026-08-20
👷 Add a weekly canary against the pinned `mockturtle` branch

### LINKED ISSUE #1124: 👷 Add a weekly canary against the pinned `mockturtle` branch
`cmake/Dependencies.cmake` pins `mockturtle` to a commit on the `mnt` branch of a fork:

```cmake
GIT_TAG c5807b7e...
```

It is the only dependency fetched as a git clone rather than a hashed tarball, and it is
the one the library depends on most deeply. Between bumps, nobody finds out whether
*fiction* still builds against that branch. Drift surfaces at the next bump, as a large
failure with an unclear cause, at the moment someone was trying to do something else.

## Work

1. Add a scheduled workflow — weekly is enough — that configures and builds against the
   `mnt` branch's current head instead of the pinned commit, and runs the test suite.
2. On failure, open or update a single tracking issue rather than opening a new one per
   run.
3. Keep it off pull requests. This job reports on an external branch, so it must never
   gate a contributor's change.

## Notes

The same pattern would suit any pinned revision that tracks a moving branch. `mockturtle`
is the one that matters; the ALGLIB and jemalloc pins are versioned releases and do not
move.

## Verification

Force the workflow against a knowingly broken revision and confirm it files an issue.
Confirm a second failure updates that issue instead of filing another.

---

Drafted with AI assistance as part of a tooling audit of this repository. The file paths, counts, and quoted configuration in this issue were verified against the working tree; the maintainer reviewed the issue slate and its scope.

marcelwa:
## Description

`cmake/Dependencies.cmake` pins `mockturtle` to a commit on the `mnt` branch of the
`marcelwa/mockturtle` fork. It is the only dependency fetched as a git clone rather than a
hashed tarball, and the one _fiction_ depends on most deeply. Between Renovate's bumps of
that pin, nothing tells us whether _fiction_ still builds against the branch, so drift
surfaces at the next bump as one large failure with an unclear cause, at the moment someone
was trying to do something else.

`mockturtle-canary.yml` builds and tests against the branch head every Monday at 04:40 UTC
and opens or updates a single tracking issue when that fails.

Closes #1124.

### How it reaches the branch head

Through `-DFETCHCONTENT_SOURCE_DIR_MOCKTURTLE`, pointed at a checkout the job makes itself.
CMake skips the download and update steps entirely for a dependency whose source directory
is named that way, so the pin in `cmake/Dependencies.cmake` and the Renovate custom manager
that matches its literal `GIT_REPOSITORY … GIT_TAG <sha>` shape both stay exactly as they
are. Nothing about the pinned build path changes.

### Why `cpp-tests-ubuntu.yml` is touched

So the canary drives the existing build recipe rather than copying it. The new
`mockturtle-ref` input places the checkout and composes the override into `EXTRA_CMAKE_ARGS`
beside the existing arm64 `-DFICTION_Z3=OFF`; empty (the default, and what every `ci.yml`
caller passes) leaves every existing job byte-identical.

One input rather than a generic `extra-cmake-args`, because a caller job that `uses:` a
reusable workflow never gets a runner: it cannot place the source tree on the machine that
configures CMake, and the absolute path is only knowable inside the called job. One input
owning both halves is the single source of truth.

Three guards come with it:

- **ccache** — a canary run restores the shared cache, where most objects stay valid, and
  never saves to it. It shares a key with the mainline matrix and runs on `main`, so its
  objects would otherwise land in the default-branch scope that every pull request restores
  from.
- **A canary that silently tested the pin** would pass forever while testing nothing. A step
  fails the run if FetchContent cloned `mockturtle` anyway.
- **`timeout-minutes: 150`** on `build_and_test`, which it never had. Roughly three times the
  slowest observed run and far below the six-hour default. The canary is the one caller
  nobody watches.

### Reporting

A `gh` run step: list the open issues, match the title exactly, then rewrite that issue's body
or create a new one. Listing beats `--search`, whose index lags behind issue creation.
`gh issue edit --body-file` touches only the body, so the run link refreshes while the title,
labels, and assignee stay put. Closing the issue is the reset signal — the next failure finds
no open match and opens a fresh one. No third-party action is involved.

That job takes a full `ubuntu-24.04` runner rather than `ubuntu-slim`. The slim image carries
`gh`, `jq`, and `sed` today, but nothing promises it always will, and this is the one job
whose failure means nobody hears that the canary broke. Ten seconds of a standard runner once
a week buys that assumption away. `ci.yml`'s change-detection and guard jobs stay on
`ubuntu-slim`; they run plain shell and the assumption does not apply to them.

`issues: write` is new to this repository. It is scoped to the one job that needs it; the
workflow keeps `permissions: {}` at the top level like every other workflow here. A
`github.ref == 'refs/heads/main'` guard keeps a `workflow_dispatch` from a topic branch from
filing a real, labelled, assigned public issue.

### Second commit: the misspelled `mockturtle` build options

Found while reading `cmake/Dependencies.cmake` for the pin. It set `MOCKTURTLE_EXAMPLES`,
`MOCKTURTLE_EXPERIMENTS`, and `MOCKTURTLE_TEST`; `mockturtle` declares none of those. Its
options are `MOCKTURTLE_BUILD_EXAMPLES`, `MOCKTURTLE_BUILD_EXPERIMENTS`, and
`MOCKTURTLE_BUILD_TESTS`. Two of the real ones default to `OFF`, which hid the misspelling —
but `MOCKTURTLE_BUILD_EXAMPLES` defaults to `ON`, so every configure carried `mockturtle`'s
13 example executables in the build graph. The same dead name was repeated in four
`CMakePresets.json` presets and in `pyproject.toml`; those five are deleted rather than
renamed, because the `FORCE` in `Dependencies.cmake` already wins and a stale copy earns a
`CMake Warning (unused-cli)` on every configure.

## Verification

- The override mechanism was verified against this `Dependencies.cmake`, not assumed: with
  `FETCHCONTENT_SOURCE_DIR_MOCKTURTLE` set, `_deps/mockturtle-src` is never created, the
  checkout's `include/` lands on the test compile lines, and a real test translation unit
  compiles and its binary passes.
- Both branches of the new assertion step were exercised: it passes with the override in
  effect and fails with the `::error::` annotation when a clone is present.
- `prek run -a` is clean.
- `zizmor --persona=pedantic` over `.github/workflows/` reports nothing new; the only finding
  in the tree is the pre-existing `stale-action-refs` on `advanced-security/filter-sarif` in
  `codeql.yml`.
- Configuring `ci-debug` reports no unused variable, no longer configures
  `_deps/mockturtle-build/examples`, and `test_graph_coloring` builds and passes.

### How to verify the reporting path after merge

The workflow cannot be dispatched before it exists on the default branch, so this is a
post-merge step:

1. On a branch off `main`, point `mockturtle-ref` at a ref that does not exist. That fails
   the checkout in seconds and costs no build.
2. `gh workflow run mockturtle-canary.yml --ref <branch>` and confirm `🐤 Report` **skips** —
   the branch guard doing its job.
3. Land that temporarily broken revision on `main`, dispatch again, and confirm the tracking
   issue appears with the run link, the `dependencies` label, and the assignee.
4. Dispatch a second time without changing anything. Confirm the issue count does not change
   and the body now carries the **new** run URL.
5. Close the issue, dispatch a third time, and confirm a fresh issue is filed.
6. Restore `mockturtle-ref: mnt` and dispatch once to confirm green.

## Known limits

- **One compiler on one operating system.** A break that only shows on Clang, arm64, macOS,
  Windows, or the Python bindings still gets through. The Renovate bump this de-risks runs
  the full fleet; the canary buys early warning, not proof. Widening it to a matrix is a
  one-line change if the weekly cost is judged worth it.
- **Infrastructure flakes file the issue too.** A hung apt mirror looks the same as a real
  break. The standing issue caps the damage at one ticket and names the flake as a possible
  cause, but the run still has to be read before acting.
- **The issue is closed by hand.** That keeps "handled" an explicit human judgement rather
  than something the workflow decides.
- **This pull request runs the full C++ fleet**, because `cpp-tests-ubuntu.yml` sits in the
  `run-cpp-tests`, `run-cpp-linter`, `run-codeql`, and `run-packaging` change-detection
  filters. The shared recipe was preferred over a duplicated one.

## Checklist:

- [x] I have added appropriate tests and documentation. No unit tests apply to a workflow;
      the mechanism is verified as described above and both branches of the new assertion
      step were exercised. Both workflows and the issue body carry the reasoning inline.
- [x] I have created/adjusted the Python bindings for any new or updated functionality. Not
      applicable; no library surface changes.
- [ ] I have made sure that all CI jobs on GitHub pass. To be confirmed once the run completes.
      The second commit removes a `CMake Warning (unused-cli)` that the first would otherwise
      have introduced.

---

Drafted with AI assistance. The `FETCHCONTENT_SOURCE_DIR_MOCKTURTLE` behavior, the removal of
the `unused-cli` warning, and both branches of the new assertion step were reproduced locally
against this tree, and the `gh`/`jq` behavior was exercised against this repository's live
issue list; the reporting path itself is verified post-merge as described above.

### DIFF
--- a/.github/mockturtle-canary-issue.md
+++ b/.github/mockturtle-canary-issue.md
@@ -1,0 +1,20 @@
+The weekly canary in `.github/workflows/mockturtle-canary.yml` builds and tests _fiction_
+against the current head of the `mnt` branch of
+[`marcelwa/mockturtle`](https://github.com/marcelwa/mockturtle) instead of the commit
+`cmake/Dependencies.cmake` pins. That run failed.
+
+- Failing run: {{ RUN_URL }}
+- Branch under test: <https://github.com/marcelwa/mockturtle/tree/mnt>
+
+Three things produce this, in falling order of likelihood:
+
+1. The branch moved in a way _fiction_ does not follow.
+2. The branch itself is broken.
+3. Neither: the run hit an infrastructure flake, such as the apt mirror hang the workflow
+   bounds with a timeout. Read the run before acting on this issue.
+
+For the first two, Renovate's next bump of the pin carries the same failure, so fixing it
+here is what unblocks that pull request.
+
+Close this issue once the canary is green again. While it stays open, further failures
+rewrite this body instead of filing another issue.
--- a/.github/workflows/cpp-tests-ubuntu.yml
+++ b/.github/workflows/cpp-tests-ubuntu.yml
@@ -11,6 +11,13 @@ on:
         description: The C++ compiler package and binary name, e.g. `g++-13`
         required: true
         type: string
+      mockturtle-ref:
+        description: >-
+          A `marcelwa/mockturtle` ref to build against instead of the revision
+          `cmake/Dependencies.cmake` pins. Empty keeps the pin.
+        required: false
+        default: ""
+        type: string
 
 concurrency:
   group: >-
@@ -28,13 +35,21 @@ env:
   Z3_VERSION: 4.13.4
   CMAKE_BUILD_PARALLEL_LEVEL: 4
   CTEST_PARALLEL_LEVEL: 4
-  # Z3 has no ARM build in `cda-tum/setup-z3`
-  EXTRA_CMAKE_ARGS: ${{ inputs.runs-on == 'ubuntu-24.04-arm' && '-DFICTION_Z3=OFF' || '' }}
+  # First argument: Z3 has no ARM build in `cda-tum/setup-z3`. Second: FetchContent adopts the
+  # tree at `FETCHCONTENT_SOURCE_DIR_MOCKTURTLE` and skips its own clone, which is how the
+  # weekly canary reaches a branch head without editing the pin in `cmake/Dependencies.cmake`.
+  EXTRA_CMAKE_ARGS: >-
+    ${{ inputs.runs-on == 'ubuntu-24.04-arm' && '-DFICTION_Z3=OFF' || '' }}
+    ${{ inputs.mockturtle-ref != ''
+    && format('-DFETCHCONTENT_SOURCE_DIR_MOCKTURTLE={0}/mockturtle', github.workspace) || '' }}
 
 jobs:
   build_and_test:
     name: 🐧 ${{ inputs.runs-on }} ${{ inputs.compiler }}
     runs-on: ${{ inputs.runs-on }}
+    # Roughly three times the slowest observed run, and far below the six-hour default. The
+    # weekly canary is the one caller nobody watches, so a hung step has to end on its own.
+    timeout-minutes: 150
     permissions:
       contents: read # required to check out the repository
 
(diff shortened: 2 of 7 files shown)


marcelwa:
Not needed.

PR #1151, opened 2026-08-21
🔬 Measure sccache without PCH against ccache with PCH on Windows

marcelwa:
Experiment arm for #1150 — not for merge. Measures sccache with `FICTION_ENABLE_PCH=OFF` against #1150's ccache + PCH + `/Z7` fix on the 4391s `🪟 windows-2025 v143` job. The two are mutually exclusive: sccache cannot cache precompiled headers at all.

Also fixes `CACHE_OPTION` being inert in `Cache.cmake`, which is why the arm could not otherwise be built.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -84,7 +84,11 @@ jobs:
         with:
           key: "codeql-ubuntu-24.04-clang++-17"
           variant: ccache
-          save: true
+          # Only `main` writes, so a pull request restores main's cache without
+          # adding an entry of its own. Every job saving 400-700 MB on every push
+          # had the repository sitting at 10.19 GB against GitHub's 10 GB limit,
+          # evicting caches faster than they could be reused.
+          save: ${{ github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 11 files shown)


marcelwa:
Closing — this arm answered the question it was built for, and the answer is that sccache is not the win here.

See the control in #1152. At equal cache warmth ccache reaches the same hit rate as sccache (84.8% vs 85.0%) and is **16% faster** on the same job, because it keeps PCH, which sccache cannot cache at all.

Two corrections this arm produced along the way, both worth keeping and both already carried into #1150:

- `SCCACHE_IDLE_TIMEOUT=0`. The default 600s shutdown fires during `ctest` between the Debug and Release builds, and the restart resets the counters. This arm's first run reported 301 compile requests where Ninja executed 602, so its 85% hit rate described half the job.
- `CACHE_OPTION` was inert in `cmake/Cache.cmake`: `find_program(... NAMES ${CACHE_OPTION_VALUES})` searched both names in list order, so asking for sccache silently got ccache. This arm could not be built at all until that was fixed.

PR #1152, opened 2026-08-22
🔬 CONTROL: ccache with equal save policy

marcelwa:
Control for the #1150 vs #1151 comparison — not for merge. #1150 saves only from `main`, #1151 saves every run; that asymmetry alone could explain 42% vs 85% hits. This is #1150 with the same save policy as #1151.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -84,7 +84,11 @@ jobs:
         with:
           key: "codeql-ubuntu-24.04-clang++-17"
           variant: ccache
-          save: true
+          # Only `main` writes, so a pull request restores main's cache without
+          # adding an entry of its own. Every job saving 400-700 MB on every push
+          # had the repository sitting at 10.19 GB against GitHub's 10 GB limit,
+          # evicting caches faster than they could be reused.
+          save: ${{ github.ref == 'refs/heads/main' }}
           max-size: 10G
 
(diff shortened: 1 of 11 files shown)


marcelwa:
Control complete — this answered the question and is not for merge.

At equal cache warmth the two approaches are not close, but not in the direction the earlier runs suggested:

| `🪟 windows-2025 ClangCL`, ~85% hits both | duration | hits |
| --- | --- | --- |
| ccache + PCH + `/Z7` (this control) | **2019s** | 512/604 (84.8%) |
| sccache, no PCH (#1151) | 2414s | 512/602 (85.0%) |

ccache and sccache reach an **identical hit rate**, so the cache technology was never the variable. At equal warmth ccache + PCH wins by 16%, because PCH still accelerates the ~15% of compilations that actually run.

The apparent 'sccache wins by 24%' in #1151 was sccache-warm measured against ccache-cold.

This control's own `v143` job makes the real problem visible from the other side: **0 of 604 hits even on its second pass**, because its ~714 MB entry did not survive while ClangCL's ~430 MB one did. Larger cache, evicted first.

Two mechanisms, both fixed in #1150: the repository sits at its 10 GB cache limit, and Actions caches are branch-scoped, so only `main`'s are visible repo-wide — every PR branch writing its own was spending the shared budget on caches almost nothing could reuse.

PR #1163, opened 2026-08-25
⬆️👷 Bump the pinned Z3 version to 4.14.1 and enable Z3 on ARM

marcelwa:
## Description

Two related commits, kept together because each one alone costs a full CI run.

### 1. Bump the pinned Z3 version to 4.14.1

The pin had drifted since December 2024 — Renovate does not manage it. Updated in ten workflows (`Z3_VERSION`) and in the `Dockerfile` (`z3-solver==`):

`codeql.yml`, `cpp-coverage.yml`, `cpp-experiments.yml`, `cpp-linter.yml`, `cpp-tests-macos.yml`, `cpp-tests-ubuntu.yml`, `cpp-tests-windows.yml`, `docstrings.yml`, `packaging-wheels.yml`, `python-tests.yml`

**Why not 5.1.0**, the actual latest — `cda-tum/setup-z3` resolves it fine, but:

| Z3 | Linux x64 asset | requires |
|---|---|---|
| 4.13.4 (before) | `x64-glibc-2.35` | `GLIBC_2.34`, `GLIBCXX_3.4.30` |
| **4.14.1 (this PR)** | `x64-glibc-2.35` | `GLIBC_2.34`, `GLIBCXX_3.4.30` |
| 4.15.0 – 5.1.0 | `x64-glibc-2.39` | `GLIBC_2.38`, `GLIBCXX_3.4.32` |

(from `objdump -T bin/libz3.so` on the downloaded release archives)

The `ubuntu-22.04` / `g++-11` entry in the `🐧 Test` matrix — deliberately the oldest supported combination — ships glibc 2.35, so anything from 4.15.0 on fails to link there. `cpp-tests-ubuntu.yml` records that ceiling next to the pin. The `z3++.h` surface *fiction* uses is unchanged between 4.13.4 and 4.14.1.

### 2. Enable Z3 on ARM

**CI.** The comment *"Z3 has no ARM build in `cda-tum/setup-z3`"* was wrong — `z3-4.14.1-arm64-glibc-2.34.zip` needs only `GLIBC_2.34` / `GLIBCXX_3.4.29`, well under what `ubuntu-24.04-arm` provides. The `-DFICTION_Z3=OFF` override and the step's `if:` guard are gone, so both ARM legs now exercise the exact solver. Dropping the hardcoded `platform: linux` / `architecture: x64` lets the action detect the host, matching what the macOS and Windows jobs already do. `EXTRA_CMAKE_ARGS` had no other writer and is removed with it.

**CD / wheels.** `pyproject.toml` already sets `FICTION_Z3 = "ON"` unconditionally — there was no ARM override to lift. The ARM wheel was nevertheless getting Z3 the slow way:

| job (run 32904441730) | `before_all` = `pip install z3-solver>=4.8.5` |
|---|---|
| `🛞 ubuntu-24.04` | **2.39 s** — installs a wheel |
| `🛞 ubuntu-24.04-arm` | **1152.82 s** — compiles Z3 from source |

Every `z3-solver` aarch64 wheel on PyPI is tagged `manylinux_2_34` or `manylinux_2_38`; none installs in cibuildwheel's default `manylinux_2_28_aarch64`, so pip falls back to the sdist. Two changes fix it:

- `manylinux-aarch64-image = "manylinux_2_34"` — the oldest image that can install one. `x86_64` stays on `manylinux_2_28`, so the [ALPHA caveat](https://github.com/pypa/manylinux/issues/1725) on `manylinux_2_34`, which is about the x86-64-v2 baseline of RHEL 9 derivatives, does not apply.
- `before-all` pins `z3-solver==4.14.1` instead of flooring at `>=4.8.5`. The pin is load-bearing, not cosmetic: pip ranks candidates by version *before* it prefers a wheel over an sdist, so a bare floor picks the newest release's sdist over an installable older wheel and source-builds regardless of the image. It also puts the wheels on the same Z3 as CI and the `Dockerfile`.

Consequence: the published `aarch64` wheel requires glibc 2.34 — Ubuntu 22.04+, Debian 12+, RHEL 9+ — which is the oldest distribution *fiction* supports in CI anyway. Verified by resolving against the full compatible tag set for each container:

```
aarch64 / manylinux_2_34 / ==4.14.1  ->  z3_solver-4.14.1.0-py3-none-manylinux_2_34_aarch64.whl
x86_64  / manylinux_2_28 / ==4.14.1  ->  z3_solver-4.14.1.0-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

`[project] dependencies` keeps the runtime floor `main` set in #1162 (`z3-solver>=4.10.2`), which is unrelated to the `before-all` pin — the former is what a user resolves against, the latter is what the wheel is built with. `🐍 Test` exercises the floor.

### Merged with `main`

#1162 landed after this branch opened and overlaps it. `python-minimums.yml` is gone, so the `Z3_VERSION` bump moved to `python-tests.yml`, which replaces it — that file already leaves `platform`/`architecture` at their `host` default, so its `ubuntu-24.04-arm` leg picks up the `arm64` archive with no further change. The changelog keeps the rewritten `**Breaking:**` bullet from `main` and restates the wheel entry against 4.10.2.

Both changes are exercised by this PR's own CI — `🐧 Test (ubuntu-24.04-arm, …)` and `🐍 Packaging (ubuntu-24.04-arm)` run here, not only on release.

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Updated the Z3 solver used across development, testing, analysis, documentation, packaging, and container environments to version 4.14.1.
  * Improved ARM-based C++ test support by enabling standard solver configuration.
  * Updated Linux ARM wheel builds and documented platform compatibility details.

* **Documentation**
  * Clarified supported solver versions, cross-platform testing, Linux wheel configuration, and ABI testing behavior.


marcelwa:
## Fixes Applied Successfully

Fixed 1 file based on 1 unresolved review comment.

**Files modified:**
- `docs/changelog.rst`

**Commit:** `5629bf31`

Applied with one deviation from the proposed wording, so flagging it here rather than silently: the suggestion kept the *"otherwise no `z3-solver` wheel is installable and the build compiles Z3 from source"* clause, which lands the entry at 340 characters against the 350 ceiling in `docs/AGENTS.md`. That file also says rationale belongs in a code comment at the site it applies to, or in the pull request — and this rationale is already in the `before-all` comment in `pyproject.toml` and in the PR description. Dropping it gets the entry to two sentences and 257 characters while keeping the explicit glibc 2.34 requirement the review asked to preserve.

The latest autofix changes are on the `bump-z3-version` branch.

marcelwa:
## Review round 2 — no changes applied

One unresolved comment, skipped as not valid. No commit this round.

**`docs/changelog.rst:186-188` — "Correct the previous floor in the changelog"** (🟡 Minor)

The suggestion is to change `raised the declared z3-solver floor from 4.8.0 to 4.10.2` so it reads `from 4.8.5`. The `4.8.0` is correct and the change would introduce an error.

`Unreleased` records the delta since the last release, and `v0.7.0` shipped `z3-solver>=4.8.0`:

```console
$ git show v0.7.0:pyproject.toml | grep z3-solver
    "z3-solver>=4.8.0"
```

4.8.5 was an intermediate value inside this same unreleased cycle — #1144 raised 4.8.0 → 4.8.5 on 2026-08-21, #1162 raised 4.8.5 → 4.10.2 on 2026-08-26, both after `v0.7.0` on 2026-07-31. No release ever declared 4.8.5, so collapsing the cycle to `4.8.0 → 4.10.2` is what the entry should say.

The `4.8.5` in the `pyproject.toml` comment refers to the bound before *that commit*, not before the last release, which is where the inference went wrong.

This entry also predates this branch — it came in with #1162 and the merge preserved it verbatim.

Leaving the thread open rather than resolving it, in case a maintainer reads the convention differently.

PR #1203, opened 2026-09-11
✨ Bind on-the-fly SiDB circuit design to Python

simon1hofmann:
## Description

Expose SiDB circuit design for placed hexagonal layouts so Python clients such as MNT Designer can use it. The binding snapshots inputs before releasing the GIL; supporting fixes isolate cloned clocking data and remove the entry point's Z3 dependency. Tested together with Designer using real SVG/SiQAD exports and browser downloads, alongside 347 Python tests.

## Checklist:

AI-assisted implementation and review; automated checks ran locally. Human approval is pending.

## Summary by CodeRabbit

- **New Features**
  - Added Python support for designing SiDB circuits from placed and routed hexagonal gate-level layouts.
  - Added configurable design policies and parameters for gate libraries, canvas options, defect influence, and crossing/wire selection.
  - Added clear handling for unsupported layouts, gates, orientations, and insufficient canvases.

- **Bug Fixes**
  - Clocked layout copies now maintain independent clocking settings.
  - On-the-fly SiDB design can compile without optional Z3 support.

- **Documentation**
  - Expanded usage, parameters, exports, errors, and defective-surface behavior documentation.


marcelwa (commented):
Overall this is in good shape: clean binding, real tests, the GIL release is genuinely safe, and the docs rewrite is an improvement over the old blacklist walkthrough. Notes below, roughly by importance.

## Correctness — things I checked and that hold up

- **GIL release is safe.** `design_gates_params` / `on_the_fly_gate_library_params` are pure value types (no `std::function`, no progress callback), so nothing re-enters Python while the GIL is released. `params` is taken by value, and `layout.clone()` snapshots before the release.
- **Throwing under a released GIL is fine here.** `release` is a local in the lambda, so unwinding out of the `catch` destroys it and reacquires the GIL before nanobind's translator runs. Worth keeping in mind if anyone later moves the `try` block or the release object.
- **`clocked_layout::clone()` fix is correct.** `clocked_layout_storage`'s implicit copy copies the `shared_ptr<clocking_scheme_t>`, so clones shared clock-number overrides; `replace_clocking_scheme(*strg->clocking)` deep-copies it. The new test in `test_clocked_layout.cpp` pins both directions, which is the right shape.
- **Z3 removal works.** `exact.hpp` self-guards on `FICTION_Z3_SOLVER` (line 21), so the unconditional include is harmless and the non-Z3 build compiles.

## Suggestions

1. **`clocked_layout.hpp:129` needs a comment.** Calling `replace_clocking_scheme` inside `clone()` reads as redundant — the storage was just copied. Its actual purpose (breaking the shared `clocking` pointer) is invisible. One line:

   ```cpp
   // the storage copy shares the clocking scheme pointer; re-seat it so clock-number
   // overrides do not propagate between clones
   ```

   Alternatively, give `clocked_layout_storage` a copy constructor that deep-copies `clocking` — that fixes it for every copy path at once rather than only for `clone()`. Worth at least considering: is there any other place that copies the storage?

2. **Move the Z3-only includes inside the guard.** `exact.hpp` and `surface_analysis.hpp` (lines 21–22) are only needed by the `#if (FICTION_Z3_SOLVER)` regions. Pulling a 3.3k-line header into every non-Z3 TU that touches circuit design costs compile time for nothing. Same for `<optional>` / `mockturtle/utils/stopwatch.hpp` if they are only used by the stats struct.

3. **`complex_gate_design_policy` is a very generic top-level Python name** for what is a nested enum of one params struct. Every other enum in the module reads as owned (`design_sidb_gates_mode`, `sidb_simulation_engine`). `sidb_complex_gate_design_policy` would age better and matches the neighbouring `sidb_on_the_fly_gate_library_params`.

4. **Duplicated Doxygen block.** `register_sidb_generators.cpp` now carries a full `@brief`/`@param` comment on the forward declaration that is verbatim identical to the definition's in `on_the_fly_circuit_design.cpp`. The other two forward declarations right above it (`random_layout_generator`, `design_gates`) have none. Drop it for consistency — the definition is the right place.

5. **Stray sentence in the Python docs.** In `on_the_fly_circuit_design.md`, "The `PRUNING_ONLY` gate-design mode skips operational simulation and returns candidates." appears in the Python section with no setup and no connection to the surrounding paragraph about exceptions and unsupported inputs. Either tie it to a sentence about choosing `design_mode`, or drop it.

6. **CI runtime of `test_design_and_export`.** This runs a real QuickCell design with 3 canvas SiDBs over an AND plus its wires and I/O — the only test here that does actual physical work. How long does it take on a CI runner? If it is more than a few seconds it deserves a marker so it does not silently inflate every Python job.

7. **Minor:** `const circuit_params params` — the top-level `const` on a by-value parameter is noise. And `test_parameters` pins exact default values (`canvas_sidb_complex_gates == 3`, `influence_radius_charged_defects == 15`); those are documented API defaults so it is defensible, but the writability assertions are the part carrying the value — the default-value half will need updating on any tuning change.

## Not blocking, just noting

The docstring promise "This function does not perform placement and routing or accept a defective surface" plus the "no timeout; use a separate process" note is exactly the right thing to say up front — good call.

include/fiction/layouts/clocked_layout.hpp
```diff
@@ -126,6 +126,7 @@ class clocked_layout : public CoordinateLayout
     {
         auto copy = clocked_layout(CoordinateLayout::clone());
         copy.strg = std::make_shared<clocked_layout_storage>(*strg);
+        copy.replace_clocking_scheme(*strg->clocking);
```
marcelwa, line 127: This line reads as redundant — the storage was just copied one line above. Its actual purpose (re-seating the `shared_ptr<clocking_scheme_t>` so clock-number overrides stop propagating between clones) is invisible to the next reader, who will be tempted to delete it.

```suggestion
        // the storage copy shares the clocking scheme pointer; re-seat it so clock-number
        // overrides do not propagate between clones
        copy.replace_clocking_scheme(*strg->clocking);
```

Worth considering the alternative: give `clocked_layout_storage` a copy constructor that deep-copies `clocking`. That fixes every copy path at once rather than only `clone()` — is there anywhere else the storage gets copied?

include/fiction/technology/sidb/generators/on_the_fly_circuit_design.hpp
```diff
@@ -71,6 +71,8 @@ class unsuccessful_gate_design_error : public std::runtime_error
      */
     explicit unsuccessful_gate_design_error(const std::string_view msg) : std::runtime_error(std::string{msg}) {}
 };
+#if (FICTION_Z3_SOLVER)
```
marcelwa, line 81: The guards do their job (`exact.hpp` self-guards on `FICTION_Z3_SOLVER` at its line 21, so the non-Z3 build compiles), but the includes at the top of this file are still unconditional. `exact.hpp` (line 21) and `surface_analysis.hpp` (line 22) are only needed inside these `#if` regions — pulling a 3.3k-line header into every non-Z3 TU that touches circuit design costs compile time for nothing. Same question for `<optional>` and `mockturtle/utils/stopwatch.hpp` if they are only used by the stats struct.

bindings/mnt/pyfiction/src/pyfiction/technology/sidb/generators/on_the_fly_circuit_design.cpp
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Choice between predefined and synthesized complex gates.
+     */
+    using complex_policy = library_params::complex_gate_design_policy;
+    /**
+     * Parameters for circuit design on a placed and routed layout.
+     */
+    using circuit_params = fiction::sidb::generators::on_the_fly_circuit_design_params;
+
+    py::enum_<complex_policy>(m, "complex_gate_design_policy",
```
marcelwa: `complex_gate_design_policy` is a very generic top-level name in `mnt.pyfiction` for what is a nested enum of one params struct. Every neighbouring enum reads as owned — `design_sidb_gates_mode`, `sidb_simulation_engine` — and the struct you register three lines down is already `sidb_on_the_fly_gate_library_params`. `sidb_complex_gate_design_policy` would match and age better.

bindings/mnt/pyfiction/src/pyfiction/technology/sidb/generators/on_the_fly_circuit_design.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    py::class_<circuit_params>(m, "on_the_fly_sidb_circuit_design_params",
+                               DOC(fiction_sidb_generators_on_the_fly_circuit_design_params))
+        .def(py::init<>(), "Default constructor.")
+        .def_rw("sidb_on_the_fly_gate_library_parameters", &circuit_params::sidb_on_the_fly_gate_library_parameters,
+                DOC(fiction_sidb_generators_on_the_fly_circuit_design_params_sidb_on_the_fly_gate_library_parameters));
+
+    m.def(
+        "on_the_fly_sidb_circuit_design",
+        [](const py_hexagonal_gate_layout& layout, const circuit_params params)
```
marcelwa: Nit: the top-level `const` on a by-value parameter is noise — drop it, or take `const circuit_params&` (nanobind copies into the caller's frame under the GIL either way, so the snapshot guarantee is unaffected).

bindings/mnt/pyfiction/src/pyfiction/technology/sidb/generators/register_sidb_generators.cpp
```diff
@@ -21,11 +21,18 @@ namespace pyfiction
 
 void random_layout_generator(nanobind::module_& m);
 void design_gates(nanobind::module_& m);
+/**
+ * @brief Registers SiDB circuit design and its gate-library parameters.
+ *
+ * @param m Python module.
+ */
+void on_the_fly_circuit_design(nanobind::module_& m);
```
marcelwa, line 24: This `@brief`/`@param` block is verbatim identical to the one on the definition in `on_the_fly_circuit_design.cpp`, and the two forward declarations directly above (`random_layout_generator`, `design_gates`) carry none. Drop it here — the definition is the right home.

docs/technology/sidb/generators/on_the_fly_circuit_design.md
```diff
@@ -1,34 +1,98 @@
...
+:::
+
+:::{tab-item} Python
+:sync: python
+
+The Python call accepts `hexagonal_gate_layout` and returns `sidb_layout`.
+The call leaves its input unchanged and releases the GIL during circuit design.
+An unsupported gate type or orientation raises `ValueError`; an unsuccessful gate search
+raises `RuntimeError`. This interface does not accept defective surfaces or logic networks.
+The `PRUNING_ONLY` gate-design mode skips operational simulation and returns candidates.
```
marcelwa: This sentence arrives with no setup and no connection to the surrounding paragraph, which is about exceptions and unsupported inputs. `PRUNING_ONLY` is not mentioned anywhere else on the page. Either tie it into a sentence about choosing `design_mode`, or drop it.

bindings/mnt/pyfiction/test/technology/sidb/generators/test_on_the_fly_circuit_design.py
```diff
@@ -0,0 +1,111 @@
...
+    library.canvas_sidb_complex_gates = 4
+    library.using_predefined_crossing_and_double_wire_if_possible = complex_gate_design_policy.DESIGN_ON_THE_FLY
+    library.influence_radius_charged_defects = 10
+    assert params.sidb_on_the_fly_gate_library_parameters.design_gate_params.number_of_canvas_sidbs == 2
+    assert library.canvas_sidb_complex_gates == 4
+    assert library.using_predefined_crossing_and_double_wire_if_possible == complex_gate_design_policy.DESIGN_ON_THE_FLY
+    assert library.influence_radius_charged_defects == 10
+
+
+def test_design_and_export(and_circuit: hexagonal_gate_layout, tmp_path: Path) -> None:
```
marcelwa, line 77: This is the only test here that does real physical work — a QuickCell design with 3 canvas SiDBs across the AND, its wires, and the I/O. How long does it take on a CI runner? If it is more than a few seconds it deserves a marker so it does not silently inflate every Python job on every run.

bindings/mnt/pyfiction/test/technology/sidb/generators/test_on_the_fly_circuit_design.py
```diff
@@ -0,0 +1,111 @@
...
+    """Return a placed AND circuit with two inputs and one output."""
+    layout = hexagonal_gate_layout((2, 2, 0), "ROW", "AND")
+    first = layout.create_pi("a", (0, 0, 0))
+    second = layout.create_pi("b", (1, 0, 0))
+    gate = layout.create_and(first, second, (1, 1, 0))
+    layout.create_po(gate, "f", (0, 2, 0))
+    return layout
+
+
+def test_parameters() -> None:
```
marcelwa, line 48: The writability half of this test is the part carrying the value. The exact default values (`canvas_sidb_complex_gates == 3`, `influence_radius_charged_defects == 15`) are documented API defaults so pinning them is defensible, but it does mean any tuning change lands as a Python test failure. Fine to keep — flagging it as a conscious choice rather than an oversight.

marcelwa (changes requested):
The substance is sound — the GIL release is genuinely safe (`design_gates_params` is a pure value type, no callbacks re-enter Python), the unwinding path reacquires the GIL before nanobind translates, and the `clocked_layout::clone()` fix correctly breaks the shared `clocking` pointer. Requesting changes for the annotated items; none of them are deep, but 1 and 2 are worth settling before merge.

bindings/mnt/pyfiction/test/technology/sidb/generators/test_on_the_fly_circuit_design.py
simon1hofmann: Kept these assertions deliberately: they check the documented API defaults, alongside the nested-field writability checks. A change to those defaults should update the assertions too.

AI-assisted reply; validation was automated.

include/fiction/layouts/clocked_layout.hpp
simon1hofmann: Added the explanation in ce2008a5. I checked every `clocked_layout_storage` use: `clone()` is the only storage-copy site; the other constructors create storage or intentionally share an existing pointer, so a custom copy constructor is unnecessary.

AI-assisted reply; validation was automated.

bindings/mnt/pyfiction/src/pyfiction/technology/sidb/generators/on_the_fly_circuit_design.cpp
simon1hofmann: Renamed the Python enum to `sidb_complex_gate_design_policy` in ce2008a5, including the package exports, tests, and docs. The C++ nested enum is unchanged.

AI-assisted reply; validation was automated.

include/fiction/technology/sidb/generators/on_the_fly_circuit_design.hpp
simon1hofmann: Moved the Z3-only project and library includes under `FICTION_Z3_SOLVER` in ce2008a5. Circuit-design instantiation and targeted include-cleaner checks pass both with and without Z3.

AI-assisted reply; validation was automated.

bindings/mnt/pyfiction/src/pyfiction/technology/sidb/generators/register_sidb_generators.cpp
simon1hofmann: Removed the duplicate declaration docblock in ce2008a5; the definition retains it.

AI-assisted reply; validation was automated.

bindings/mnt/pyfiction/src/pyfiction/technology/sidb/generators/on_the_fly_circuit_design.cpp
simon1hofmann: Dropped the top-level `const` in ce2008a5. I kept the by-value parameter: nanobind's bound-class caster can return a reference to the Python-owned C++ instance, so `const circuit_params&` would not guarantee an independent snapshot while the GIL is released.

AI-assisted reply; validation was automated.

bindings/mnt/pyfiction/test/technology/sidb/generators/test_on_the_fly_circuit_design.py
simon1hofmann: The previous CI run did not record per-test durations, so its whole-suite totals cannot answer this reliably; the rebuilt local test took 9.23 s. ce2008a5 adds a `slow` marker and `--durations=10`: `pytest -m 'not slow'` skips it for quick runs, while default CI and wheel tests retain physical integration coverage and report timings.

AI-assisted reply; validation was automated.

docs/technology/sidb/generators/on_the_fly_circuit_design.md
simon1hofmann: Connected the caveat to selecting `design_gate_params.design_mode` in ce2008a5. The text explicitly says that `PRUNING_ONLY` skips operational simulation and does not verify the designed gates' functionality.

AI-assisted reply; validation was automated.

bindings/mnt/pyfiction/test/technology/sidb/generators/test_on_the_fly_circuit_design.py
simon1hofmann: At ce2008a5, `test_design_and_export` has these pytest `call` durations across the minimum-dependency run on Python 3.10 and tests on Python 3.10–3.14:

| CI runner | Duration range |
|---|---:|
| [Ubuntu x64](https://github.com/cda-tum/fiction/actions/runs/34588229954/job/103227386353) | 19.83–21.35 s |
| [Ubuntu ARM](https://github.com/cda-tum/fiction/actions/runs/34588229954/job/103227386606) | 10.64–10.70 s |
| [macOS](https://github.com/cda-tum/fiction/actions/runs/34588229954/job/103227386379) | 8.10–14.23 s |
| [Windows](https://github.com/cda-tum/fiction/actions/runs/34588229954/job/103227386748) | 25.74–28.13 s |

All 24 executions passed; these timings cover the individual test body, not the complete suite or job, and are separate from the 9.23-second local measurement.

AI-assisted reply; timings were extracted from GitHub Actions logs.

marcelwa (approved):
Thanks for iterating this so quickly. Looks good to me now 🙂

